PR #3, opened 2014-05-16
B3462 fix namespace issue gcc 447

MattClarkson:
Hi there,

I needed these commits to compile on Scientific Linux, using 
"g++ (GCC) 4.4.7 20120313 (Red Hat 4.4.7-4)"

Thanks

Matt


SimonRit:
Thanks for the fix, I've pushed it right now.
Simon

PR #4, opened 2014-05-31
Remove unnecessary New() and typedef of FDK recon filters

Hakkk2002:
In the original code all three FDK reconstruction filters are constructed although only one will be used.
If 'cpu' is used, the CPU FDK filter is constructed twice and FDKCPUType are also defined twice.


SimonRit:
Thanks! This is really a minor change but it doesn't hurt indeed,
Simon

PR #5, opened 2014-06-06
Reduce memory used by ProjectionsReader

Hakkk2002:
After raw data are converted to attenuation, the raw data are not needed
anymore in the mini-pipeline and can be released. This can save memory
especially when --lowmem is not used in reconstruction.

Discussion: this may cause multiple unnecessary deallocation/allocation if
--lowmem is used. Maybe it is better to let SetProjectionsReaderFromGgo
determine whether to set the flag based on whether --lowmem exists?

### DIFF
--- a/code/rtkProjectionsReader.txx
+++ b/code/rtkProjectionsReader.txx
@@ -210,6 +210,10 @@ void ProjectionsReader<TOutputImage>
     m_ImageIO = imageIO;
     }
 
+  // Release output data of m_RawDataReader if conversion occurs
+  if ( m_RawDataReader != m_RawToProjectionsFilter )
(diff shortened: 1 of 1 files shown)


SimonRit:
Thanks. By the way, I don't think I know you. Where are you from? What are
you working on?

Hakkk2002:
Hi this is Chao Wu, a post-doc from TU Delft, the Netherlands.
I am partly working on CT reconstruction algorithm/software. RTK serves me
a great foundation.

2014-06-06 14:50 GMT+02:00 Simon Rit notifications@github.com:

SimonRit:
I'm glad to hear it. Continue the good work!
Simon

PR #7, opened 2014-06-11
Free GPU mem after cudaPadding

Hakkk2002:
Projections as input images are not needed in GPU after padding.

To discuss: 
GPU memory is scarce resource so it is recommended to release any unnecessary buffer immediately.
But this implementation needs to modify CudaDataManager. Please see if this scenario is acceptable or if you have better proposal.

Regards, Chao

### DIFF
--- a/code/rtkCudaFFTRampImageFilter.cxx
+++ b/code/rtkCudaFFTRampImageFilter.cxx
@@ -72,6 +72,7 @@ rtk::CudaFFTRampImageFilter::PadInputImageRegion(const RegionType &inputRegion)
                pout,
                m_TruncationMirrorWeights );
 
+  this->GetInput()->GetCudaDataManager()->ForceReleaseGPUBuffer();
   return paddedImage.GetPointer();
 }
 
--- a/utilities/ITKCudaCommon/include/itkCudaDataManager.h
+++ b/utilities/ITKCudaCommon/include/itkCudaDataManager.h
@@ -49,6 +49,14 @@ public:
     CUDA_CHECK(cudaMalloc(&m_GPUBuffer, bufferSize));
     }
 
+  void Release()
+    {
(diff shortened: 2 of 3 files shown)


SimonRit:
Hi Chao,
No, this is not acceptable. This looks good in a very specific context but
there are contexts where it is not acceptable. For example, suppose that
the input is used by another gpu filter in some other parts of the code,
then you have to re-transfer. I don't hthink that we should see the GPU
memory as something as scarce as your modifications suggest it... How much
GPU mem do you have?
Simon

Hakkk2002:
Hi Simon,

Thanks for the comment.
I have 3G GPU mem and I am dealing with things like 1944x1536 sized
projections and 1500x1500x1500 sized volumes.
The volume itself is already out-of-memory so --division has to be used.
However to make reconstruction as fast as possible, the number of divisions
should be minimized and therefore the GPU memory for the projection chunk,
padded chunk, fft image and fft plan is again restricted.
I must admit that this restriction can be relaxed by using --subsetsize and
its influence on timing is not as big as --division; actually quite small.
However my proposal is meant for an optimal use of GPU memory. In this
sense --subsetsize is a workaround not a best solution.
As you said the GPU buffer should remain in the case that other GPU filter
will use it.
So I made this force-release as a separate method of the CudaDataManager
and must be explicitly called; the calling function should be fully aware
that whether this GPU buffer would be used elsewhere.
But you are right that in this case the GPU buffer is for an input object
of a filter, thus whether it will be used again can only be determined with
context not within the scope of filter itself.
So I'll keep it for my own use at the moment. If I have a better scenario
later I'll come back to this.

Regards,
Chao

2014-06-11 15:39 GMT+02:00 Simon Rit notifications@github.com:

SimonRit:
Ok, huge data indeed. In this case, I would consider leaving ramp filtering
to the CPU with FFTW. This is quite efficient already provided that you
have a good computer and it releases some memory for the volume. The other
option is to buy more than one GPU or a GPU with more RAM...
Good luck,
Simon

PR #10, opened 2014-10-17
Added Ram-Lak and SheppLogan window function in Ramp filter.

ipsusila:
Dear Simon Rit,
I currently using RTK for reconstruction of industrial CT data. I add Ram-Lak and Shepp-Logan to the Ramp filter. I was considering to separate the ramp filter class and window function class (to easily add new/custom window function in the future), but it will affects the current API, so I end up with this solution. Please consider to merge these changes.

best regards,
P.S. I think I met you during ICCR 2013 in Melbourne.

### DIFF
--- a/code/rtkFFTRampImageFilter.h
+++ b/code/rtkFFTRampImageFilter.h
@@ -130,11 +130,23 @@ public:
   itkGetConstMacro(HannCutFrequencyY, double);
   itkSetMacro(HannCutFrequencyY, double);
   
-  /** Set/Get the Ram-Lak window frequency (0...1). 0 (default) disable it */
+  /** Set/Get the Ram-Lak window frequency (0...1). 0 (default) disable it.
+   * Equation and further explanation about Ram-Lak filter could be found in:
+   * 1. Fundamentals of 2D and 3D reconstruction (by Dr. Günter Lauritsch). 
+   *    http://campar.in.tum.de/twiki/pub/Chair/TeachingWs04IOIV/08CTReconstruction.pdf
+   * 2. Reconstruction. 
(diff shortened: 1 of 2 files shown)


SimonRit:
Thank you this is great. Could you add a link to some references in the doxygen to make it clear? It can either be the original publication or other links on the web.
Maybe we should add a wiki page with a description of each filter to make all this clearer because it's beginning to be quite messy... If you want to suggest a text and a figure ploting them, I'll add it.
Thanks again for sharing the code,
Simon

ipsusila:
Thank your for your consideration. I got the reference from these URLs:
http://campar.in.tum.de/twiki/pub/Chair/TeachingWs04IOIV/08CTReconstruction.pdf
http://oftankonyv.reak.bme.hu/tiki-index.php?page=Reconstruction
http://www.kps.or.kr/jkps/downloadPdf.asp?articleuid=%7B8EC731EB-D625-45F8-B73B-9815B63656E1%7D

I will add them as doxygen comment. I also make a plot of "ramp kernel amplitude vs frequency" calculated in FFTRampImageFilter::GetFFTRampKernel, and put the result in https://github.com/ipsusila/learning/blob/master/ct/ramp-filter.png 

ramp-filter

Yes, I agree with you about the wiki. Were and how we can we add the wiki? I also agree with this --> "it's beginning to be quite messy". This is the reason why I consider to refactor FFTRampImageFilter and WindowFunction into separate class. After cleaning up the code, I will share it. 

I Putu Susila

SimonRit:
Hi,
Could you provide a username and an email address for the wiki? I will then
create a user account and you can write the wiki page.
For the code, would you prefer to merge your changes now or to wait for the
new version?
Thanks again,
Simon

PR #14, opened 2015-07-24
Allowing several warp_ggo in the same CMakeLists.txt

dsarrut:
Initialize the list of ggo files for a new warp_ggo command. 

It allows to put several warp_ggo in the same CMakeLists.txt.


SimonRit:
Merci !

PR #19, opened 2016-02-12
BugFix: changed AddProjections to AddProjectionsInRadians in rtkDigisensGeometryReader.cxx

cdeepakroy:
AddProjection() expects all angles to be in degrees. The [code](https://github.com/SimonRit/RTK/pull/19/files#diff-3fd809ed536af9fbddb42d5b75126856R114) in rtkDigisensGeometryReader.cxx erroneously supplied angles in radians to this function.

@SimonRit If you think this fix is ok, i can modify the test file on MIDAS as well to make the test pass.

### DIFF
--- a/code/rtkDigisensGeometryReader.cxx
+++ b/code/rtkDigisensGeometryReader.cxx
@@ -111,14 +111,14 @@ rtk::DigisensGeometryReader
     ThreeDTransformType::Pointer xfm3D = ThreeDTransformType::New();
     xfm3D->SetMatrix(xfm3DVersor.GetMatrix());
 
-    m_Geometry->AddProjection( sid,
-                               sdd,
-                               xfm3D->GetAngleY(),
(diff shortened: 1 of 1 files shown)


SimonRit:
Sorry about that... Yes please, send a new test file by email and I'll merge the whole thing when the tests will pass.
Thanks a lot for your contribution!

SimonRit:
Hi Deepak, what's the status of this contribution? Do you confirm that the angles are in radians and should I just update the baseline file for the test?

cdeepakroy:
@SimonRit sorry for letting this linger for so long. Yes the angles are in radians and the change from the call to AddProjection and AddProjectionInRadians is correct. We just need to change the test file to make it pass. Please do it if you can. If not, I will try to do it this weekend.

SimonRit:
Done, thanks again for the contribution.
Simon

PR #24, opened 2016-05-25
Reorderprojectionsandgeometry

cyrilmory:
In 4D applications, sorts the projections and the geometry by ascending phase, so that adjacent projections may have the same phase, which opens the way to optimization (by processing  projections in slabs instead one by one).
This pull request therefore also contains a re-writing of rtkCudaForwardProjectionImageFilter and rtkCudaBackProjectionImageFilter that implements this optimization.


SimonRit:
Thanks, quite some work. I had a first glance at it. Can you explain why you removed kernel_optim in rtkCudaBackProjectionImageFilter? Is this optim useless with respect to the new one? I guess we also need to propagate your changes to FDK backprojection.

cyrilmory:
I commented out kernel_optim only for development efficiency reasons, in order to have only one code to modify until everything starts working. I am planning on uncommenting it and making it consistent with the rest (I have already started working on it). I have doubts on the speedup gained with this optimization, but it is not much work to keep and update it. We can decide later whether it is relevant to completely remove it. 

Indeed, the changes should be propagated to CudaFDKBackProjection (I'm also on it).

cyrilmory:
My bad, it is actually more complex than I thought. Kernel_optim seems to run less threads, with each thread updating several voxels one after the other. it seems incompatible with the optimization I implemented, in which each thread updates only one voxel, accumulating the contributions from each projection in a fast-to-access buffer, and then writing to slow global memory only once. 
Since kernel_optim spares calculations, while processing several projections at once spares write accesses in global memory, I think processing several projections will be faster. I would remove kernel_optim entirely.

dchansen:
Out of curiosity, why are you using a 3D texture for the projections, rather than a 2D layered texture?

cyrilmory:
@dchansen : thanks for the suggestion, I didn't know about 2D layered textures. I just looked into it, I'll try it :)

PR #27, opened 2016-05-28
Polygaincorrection

sbrousmiche:
Add a polynomial gain correction filter (cpu + cuda)
Add associated application and test


SimonRit:
I have corrected it (next time you do those corrections) and merged. However, I don't think anyone else can use it without an how to. Do you have the data to do a wiki page? That would be great!
Thanks,
Simon

PR #44, opened 2016-07-19
BUG: Requested BoundaryCondition was not respected

### LINKED ISSUE #43: BoundaryCondition override is not respected for rtkBackwardDifferenceDivergenceImageFilter.hxx
The BoundaryCondition override is not using the provided boundary condition.

hjmjohnson:
When choosing an alternate boundary condition (i.e.
PeriodicBoundary), it was not being used.

This patch resolves #43.


hjmjohnson:
@SimonRit @gregsharp 

This appears to be a minor oversight for implementing alternate boundary conditions.

Thank you for the nice package of tools.  I am using some of these algorithms for other (non cone-beam) reconstruction needs.

Thanks,
Hans

SimonRit:
Yes, tests still pass so it's probably an oversight. Thanks,
Simon

PR #45, opened 2016-07-19
STYLE: Prefer initialization to assignment & local variables

hjmjohnson:
During a code review of this function, minor style changes to
ease code readability were introduced.  Prefer initialization to
assignment, and prefer that variables have as small of scope as possible.


SimonRit:
Thanks!

PR #47, opened 2016-07-29
PERF: Prefer multiplication to division

hjmjohnson:
Multiplication is a faster operation than division, so
prefer to store 1/spacing values to be used repeatedly.

### DIFF
--- a/code/rtkBackwardDifferenceDivergenceImageFilter.hxx
+++ b/code/rtkBackwardDifferenceDivergenceImageFilter.hxx
@@ -137,8 +137,18 @@ void
 BackwardDifferenceDivergenceImageFilter< TInputImage, TOutputImage>
 ::BeforeThreadedGenerateData()
 {
-  m_SpacingCoeffs= this->GetInput()->GetSpacing();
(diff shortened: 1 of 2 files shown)


code/rtkBackwardDifferenceDivergenceImageFilter.hxx
```diff
@@ -137,8 +137,18 @@ void
...
 ::BeforeThreadedGenerateData()
 {
-  m_SpacingCoeffs= this->GetInput()->GetSpacing();
-  if (m_UseImageSpacing == false) m_SpacingCoeffs.Fill(1);
+  if (m_UseImageSpacing == false)
+    {
+    m_InvSpacingCoeffs.Fill(1.0);
+    }
+  else
+  {
```
nocnokneo: Brace should be indented.

hjmjohnson: @nocnokneo Fixed.  Thanks.  It should be ready to merge.

hjmjohnson:
@nocnokneo Comments addressed.

SimonRit:
Makes sense, thanks.

PR #48, opened 2016-07-29
BUG: Pixel accumulator type should match image pixel type

hjmjohnson:
Prefer initialization to assignment, and make types match
to avoid typecasting.


SimonRit:
This is something I don't know: what is the difference between 0.0F and 0.0f? I'm merging anyway, it seems to compile fine. Thanks.

PR #51, opened 2016-07-29
Cleanup c make

hjmjohnson:


nocnokneo:
+1

hjmjohnson:
@nocnokneo This should be ready to merge now.  It would be nice to clear a few of these items so that they don't stack up.

SimonRit:
I'm not a big fan of these large changes but I guess it's good to get up-to-date with ITK standard since RTK is meant to be ITK-like. Thanks for the effort.

PR #53, opened 2016-08-16
COMP: Conform header include guard names to ITK style

hjmjohnson:
Instead of __rtkFoo_h now expect rtkFoo_h

Double underscore prefixes are reserved in C++ and so should not be used.
Fixes new clang warning -Wreserved-id-macro.

find . ( -iname _.h -and ! -path *utilities_ ) -exec perl -pi -w -e 's/__rtk(.*)_h/rtk$1_h/g;' {} \;

This brings RTK into compliance with ITK standards introduced Dec 15, 2014 by ITK.
This also conforms to best practices where double
underscores "__" are reserved for compiler reserved id macros.


hjmjohnson:
@SimonRit Fixed kwstyle issues here by matching the new ITK kwstyle guide.

PR #54, opened 2016-08-16
Allow override of output pixel type

hjmjohnson:
This provides consistency that allows better interactions with SimpleITK.

### DIFF
--- a/code/rtkForwardDifferenceGradientImageFilter.h
+++ b/code/rtkForwardDifferenceGradientImageFilter.h
@@ -15,25 +15,28 @@
  *  limitations under the License.
  *
  *=========================================================================*/
-
-#ifndef __rtkForwardDifferenceGradientImageFilter_h
-#define __rtkForwardDifferenceGradientImageFilter_h
+#ifndef rtkForwardDifferenceGradientImageFilter_h
+#define rtkForwardDifferenceGradientImageFilter_h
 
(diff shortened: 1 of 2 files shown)


SimonRit:
Hi,
We had a closer look at your pull-request. I disagree with the "using namespace", following the policy of ITK:
http://public.kitware.com/pipermail/insight-users/2009-December/034500.html
In general, please check the travis-ci log, we won't merge until there is no problem.
One suggestion since you seem to be in the ITK developer team: don't you think that this filter should be moved to ITK?
Thanks again for your contribution,
Simon

hjmjohnson:
@SimonRit   I do agree that this filter should be moved to ITK proper.  I have that on the TODO list for a graduate student that is just starting this week in my lab.

If you don't mind, can this pull request remain as pull request for a few more weeks.  I need to get my semester classes started off on the right foot before returning to addressing these issues more completely.

SimonRit:
Sure, no hurry. Thanks!

SimonRit:
Sure, no hurry. Thanks!

SimonRit:
Hi,
Any update on this pull-request? Should I close it?
Thanks,
Simon

hjmjohnson:
@SimonRit Thank you for closing.  We may not return to this task soon, so it is best to close.  I'll record it as a long-term todo (i.e. one of those that rarely gets done).

PR #72, opened 2016-11-25
Xim reader branch

agravgaard:
Addition of reader for Varian Xim projections.
The reader has only been tested on a limited data-set.
The scanning direction may need to be corrected for manually as it neither appears in the Scan.xml or the Xim header. 
Hotfix:
modify "m_Geometry->AddProjection(sid, sdd, 180.0 - angle, offsetx, offsety);"
to "m_Geometry->AddProjection(sid, sdd, angle + 180.0, offsetx, offsety);"
Also it is unclear (due to small values in the tested data-set) if offsetx and offsety will have the intended effect.


SimonRit:
Hi,
Thanks a lot! Indeed, the style check does not pass. I'll look at your pull request next week and merge it. I hope I still have the files that you had sent on the mailing list to write a test.
Simon

PR #79, opened 2017-01-23
Cylindrical voxel based

cyrilmory:


cyrilmory:
The cylindrical voxel based back projector does not produce the correct back projection results currently. Closing the pull request until this issue is fixed, to avoid unnecessary integration tests

PR #104, opened 2017-07-07
Joseph by reference

cyrilmory:


cyrilmory:
The last commit actually slows down the forward projection. Not merging it.

PR #117, opened 2017-11-20
RTK does not create proper files to export / install Targets

dkuegler:
I tried to fix most of the issues. At least for me and my students this works better to include in the environment.
The main benefit is, that I can deploy the "install" folder to students, that want/need to use RTK. So they do not need to individually build RTK.
This strategy has proven successful for many other libraries (ITK, VTK, PlusLib, etc) for us.

David


SimonRit:
Hi,
Thanks. I do not know much about these targets and I would tend to trust you. However, your PR does not seem to pass TravisCI tests. Can you have a look and fix it?
BTW, there is a new branch called [RTK-ExternalModule](/SimonRit/RTK/tree/RTK-ExternalModule) which will be a big change and make RTK an external module of ITK. Hopefully, these targets will be better defined then.
Simon

jjomier:
Looks like the following line is causing the issue when not packaging:

  include("${RTK_PACKAGE_DIR}/@RTK_EXPORTS_FILE@")

dkuegler:
So this build currently does not work, since the Targets file, which is the modern way to export cmake targets is only created in the generate step of cmake, but required for the configure step.

More expicitly: 
config RTK includes the code in examples including the lines "find_package(RTK)" and "include(${RTK_USE_FILE})" which conflicts with CMake Best practice.
The use file is supposed to initialize the cmake environmant (including the targets), but the environment (i.e. targets) are already available from their definition from within the original RTK code.

Solutions:
A) drop those two lines from the examples (edit: using the environment as already configured)
OR
B) instead of defining the examples as "subdirectories" of RTK, define it as independent projects, which are either build by cmake in the generation step (including them into the MS Projects would have to be "Manual" through [include_external_msproject](https://cmake.org/cmake/help/v3.9/command/include_external_msproject.html?highlight=project#command:include_external_msproject)) or independently in travis etc.

jjomier:
I would vote for option A which is a little bit cleaner IMHO.

dkuegler:
Option A is much less work.

dkuegler:
So i inserted a condition into the examples, which limits builds to the case where the examples ony search for RTK (find_package(RTK), if the target RTK is not defined (yet))

code/CMakeLists.txt
```diff
@@ -154,22 +157,38 @@ if (RTK_USE_CUDA)
               ${rtkcuda_CUDA_WRAPPERS}
               ${rtkcuda_CUDA_FILES})
   set_target_properties (rtkcuda PROPERTIES LINKER_LANGUAGE CXX)
-  target_link_libraries(rtkcuda ${CUDA_LIBRARIES} ${CUDA_cufft_LIBRARY} ${CUDA_cublas_LIBRARY} RTK ITKCudaCommon)
+  target_link_libraries(rtkcuda $<BUILD_INTERFACE:${CUDA_LIBRARIES}> $<BUILD_INTERFACE:${CUDA_cufft_LIBRARY}> $<BUILD_INTERFACE:${CUDA_cublas_LIBRARY}> RTK ITKCudaCommon)
```
SimonRit, line 160: This PR still does not pass TravisCI and I can reproduce the problem on my box. If I rollback this line to the previous version, it compiles fine. I don't really understand this but would someone have a fix? Are the other BUILD_INTERFACE and INSTALL_INTERFACE correct?

dkuegler: I cannot recreate the problem on my windows machine.
The build works fine.

Scanning dependencies of target HelloWorld
make[2]: *** No rule to make target `/usr/lib/x86_64-linux-gnu/librt.so>', needed by `bin/HelloWorld'.
[ 96%] Building CXX object examples/HelloWorld/CMakeFiles/HelloWorld.dir/HelloWorld.cxx.o
[ 72%] Building CXX object Code/BasicFilters/src/CMakeFiles/SimpleRTKBasicFilters2.dir/srtkIntensityWindowingImageFilter.cxx.o
make[2]: Target `examples/HelloWorld/CMakeFiles/HelloWorld.dir/build' not remade because of errors.
make[1]: *** [examples/HelloWorld/CMakeFiles/HelloWorld.dir/all] Error 2
Scanning dependencies of target FirstReconstruction
make[2]: *** No rule to make target `/usr/lib/x86_64-linux-gnu/librt.so>', needed by `bin/FirstReconstruction'.
[ 96%] Building CXX object examples/FirstReconstruction/CMakeFiles/FirstReconstruction.dir/FirstReconstruction.cxx.o
[ 72%] Building CXX object Code/BasicFilters/src/CMakeFiles/SimpleRTKBasicFilters2.dir/srtkJosephForwardProjectionImageFilter.cxx.o
make[2]: Target `examples/FirstReconstruction/CMakeFiles/FirstReconstruction.dir/build' not remade because of errors.
make[1]: *** [examples/FirstReconstruction/CMakeFiles/FirstReconstruction.dir/all] Error 2

So this looks like it tries to link against "rt" instead of "RTK" or "rtkcuda".

This does not happen on my machine.

dkuegler:
Can you check, if this is the based on the same error as https://gitlab.kitware.com/cmake/cmake/commit/1aa5c1bec82537ccbff8e05ae989b8befbdd286e 
So would updating FindCUDA solve the issue?

code/CMakeLists.txt
dkuegler: A rollback does not really work, because in that case absolute files are stored in a "distributed" file.

SimonRit:
Commit fcf906c should fix the problem, I'll merge your PR as soon as Travis CI confirms it.

PR #123, opened 2017-11-30
Rtk external module

LucasGandel:


LucasGandel:
@SimonRit  Here are some fixes for template definition when wrapping RTK for python.
The warning about "const itk::Image" is still here though.

LucasGandel:
@SimonRit I update the files to fix Python wrapping. The last commit is reverting the use of the shadow class for the writer. Because of swig behavior with ofstream, we have to shadow those function anyway.
I'll focus on continuous integration with python testing and we can come back to this later once everything is tested. I confirm that users don't need to bother with the shadow class and can keep using rtkThreeDCircularProjectionGeometryXMLFileWriter as usual.

PR #125, opened 2017-12-11
Setup style hooks

LucasGandel:
Add git hooks to check for basic style rules on every new commit.
This feature is disabled by default. To enable it run the SetupHooks.sh
script from any directory.
It will download the hooks scripts that are stored on the "hooks"
branch of the project. To add new hooks, just push the corresponding
executable hook script on that branch. (e.g. pre-push)
This commit only adds a pre-commit hook. It is run when "git commit" is
called and will run on every file specified in the .gitattributes file.
It internally calls pre-commit-style.bash to setup KWStyle checks.
RTK.kws.xml is added to specify KWStyle rules.
RTKOverwrite.txt is added to specify exception rules.
KWStyle generates temporary files for failing checks, thus we add a
.gitignore file to avoid commiting temporary files that are
in the source directory.

### DIFF
--- a/cmake/Hooks/SetupHooks.sh
+++ b/cmake/Hooks/SetupHooks.sh
@@ -1,0 +1,76 @@
+#!/usr/bin/env bash
+#==========================================================================
+#
+#   Copyright Insight Software Consortium
+#
+#   Licensed under the Apache License, Version 2.0 (the "License");
+#   you may not use this file except in compliance with the License.
+#   You may obtain a copy of the License at
+#
+#          http://www.apache.org/licenses/LICENSE-2.0.txt
+#
+#   Unless required by applicable law or agreed to in writing, software
+#   distributed under the License is distributed on an "AS IS" BASIS,
+#   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#   See the License for the specific language governing permissions and
+#   limitations under the License.
+#
+#==========================================================================*/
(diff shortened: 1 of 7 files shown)


LucasGandel:
@SimonRit Here is a branch introducing Git hooks. It needs a "hooks" branch on the repo as it is done for the "dashboard" branch.
I cannot request merge of my [hooks branch](https://github.com/LucasGandel/RTK-1/tree/hooks) because it has no history.
Could you please copy my hooks branch and add it to the repo?
Once this is done we'll just have to change the remote repo l.67 in SetupHooks.cmake.

cmake/Hooks/SetupHooks.sh
```diff
@@ -0,0 +1,82 @@
...
+fi
+
+# Grab the hooks.
+# Use the local hooks if possible.
+echo "Pulling the hooks..."
+if GIT_DIR=.. git for-each-ref refs/remotes/origin/hooks 2>/dev/null | \
+  egrep-q 'refs/remotes/origin/hooks$'; then
+  git fetch .. remotes/origin/hooks
+else
+  git fetch https://github.com/LucasGandel/RTK-1.git hooks
```
LucasGandel: @SimonRit  To be changed when the "hooks" branch is added to the repo

SimonRit:
I'm merging this pull request but we need to decide whether we keep the .gitignore file in the future.

PR #126, opened 2017-12-11
App veyor setup

LucasGandel:
Setup continuous integration with AppVeyor.


LucasGandel:
@SimonRit Here the PR for AppVeyor continuous integration. Please do not merge as it is tagged as WIP. I need to check for some weird caching behavior but then we should be all set.
The configuration file uses ctest to build and test and thus reports build warning and failure as expected. We should use the same kind of configuration for CircleCI and Travis in a near future.

LucasGandel:
Ready to be merged on my side. It is still using my appveyor account.
@SimonRit  If you create and setup your own account I can update the badge link in the README file. Otherwise you can integrate this and make the change later.

PR #130, opened 2017-12-31
COMP: Fixed logic to support ITKv5

hjmjohnson:
With ITK version 5.0.0 the ITK_VERSION_MINOR was 0, thus
causing the logic to fail.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -172,10 +172,10 @@ endif()
 list(APPEND RTK_INCLUDE_DIRS "${PROJECT_BINARY_DIR}")
 include_directories(${RTK_INCLUDE_DIRS})
 
-if(ITK_VERSION_MINOR LESS "4")
+if(ITK_VERSION LESS "4.4")
   include_directories(BEFORE utilities/itkImageScanlineConstIterator)
 endif()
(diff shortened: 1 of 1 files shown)


SimonRit:
Thanks!

hjmjohnson:
@SimonRit,

I am currently preparing for ITKv5. (C++11 only support), and RTK is in my testbed of tools to use to ensure that we don't break too much.

While I am testing, I could easily make common improvements to C++11, but those changes are going to require a more modern compilation environment. (c++11 support, and CMake 3.8.2).

For ITKv5 we have determined that the improvements from these modern development environments are worth removing support for older compilers and cmake paradigms.

Are you interested in reviewing changes that provide modernization (removing support for older versions of ITK and older versions of CMake, and updating to C++11 only support)?

Thanks,
Hans

SimonRit:
@hjmjohnson

Of course I'm interested, thanks. We (with @LucasGandel) are currently preparing a large refactoring of the RTK repository, see the [RTK-ExternalModule](/SimonRit/RTK/tree/RTK-ExternalModule) branch. I would probably merge your changes to this branch instead of the current master branch which should be released soon.

PR #133, opened 2018-01-02
Add itkoverride

hjmjohnson:
This is all the changes necessary to make a clean build with proposed ITKv5 updates.  It should be backwards compatible with recent ITKv4 as well.


hjmjohnson:
Moved to different refactoring branch.

PR #134, opened 2018-01-04
FIX: RTK did not build with ITK 4.13

dkuegler:
Build threw a couple of errors when using CMAKE 3.8.1 since utilities from ITK where included multiple times.

fixed: LESS --> VERSION_LESS


SimonRit:
Thanks

PR #135, opened 2018-01-05
ITK v5

hjmjohnson:


hjmjohnson:
@SimonRit I rebased these on top of the RTK-ExternalModule.  These now build  upon that branch.

hjmjohnson:
@SimonRit FYI:  there are more compiler warnings, but this resolves most of them.

We are working on ITKv5 hard right now.  It will be great to have RTK as an ITK external module when we announce the new release late this spring.

itk-module.cmake
```diff
@@ -20,6 +21,7 @@ set(RTK_DEPENDS
   ITKSmoothing
   ITKImageNoise
   ${RTK_IO_DEPENDS}
+  ${ITK_MODULES_ENABLED}
```
LucasGandel, line 24: This adds useless dependency on every enabled ITK module. This is required because all factories are registered when they should not, thus adding dependency on every IO module. 
IMO the right fix is to register only factories that are needed by RTK. This is done by requesting only RTK requested modules as COMPONENTS in the find_package() call. See commit 6944b0d2b8de199cf2e00ab182080ea53f95d8b8

LucasGandel:
@hjmjohnson Thank you for your changes
PR #136 was merged in preparation for ITKv5, fixing IO dependencies and rtkWarpImageFilter const correctness.
Your fix of const correctness seems cleaner. Please see my comment on IO modules though.

hjmjohnson:
@LucasGandel Your recommendation was implemented.

hjmjohnson:
@LucasGandel @SimonRit 

Can this set of fixes be incorporated?  Do you need/want me to cherry-pick some of them off into smaller requests?

LucasGandel:
@hjmjohnson Sorry for not taking a look at it earlier. Yes, that would be great to split this change into multiple MRs if you can. I can think of three different changes : 
- unused typedefs + unused variables
- ITK_OVERRIDE + const correctness
- A more global one changing the library name, cmake version ...

I would remove 45c0361, following my last comment I think 6944b0d is enough.

hjmjohnson:
Broken apart into #138 , #155, #156 , #157, #159,  in the hopes that these can be reviewed more rapidly for incorporation.

The external tool that depends on RTK has been disabled for months due to the excessive overhead needed to maintain private branches.  If this burden persists, we will need to evaluate if we can continue to depend on RTK, or if we will need to permanently fork and diverge from the main tree.

SimonRit:
We'll do our best, sorry for the delay and thanks a lot for the work!

PR #138, opened 2018-01-10
Fix find module library name

hjmjohnson:


hjmjohnson:
Need to change the name of the library.

SimonRit:
Thanks but why is this a need?

hjmjohnson:
I'm building a tool that depends on RTK.  That tool uses "find_package(RTK)" to identify and use the "RTK_LIBRARIES" variable to link against.

hjmjohnson:
The library name that is currently generated is "libitkRTK-5.0.a" when built against ITK v5, and "libitkRTK-4.13.a" when built against ITK 4.13.

Hans

hjmjohnson:
Can this please be incorporated?

The find_package fails without this patch because the name of the library that is searched for is wrong.

Hans

SimonRit:
@LucasGandel There is indeed a problem with the library name. Currently, on RTK-ExternalModule or with this patch, I cannot compile on MacOS with dynamic libraries. Can you try to fix this?

LucasGandel:
@SimonRit I'll try to reproduce on MacOS and have a look at AppVeyor failures.

hjmjohnson:
Addressed in #165

PR #141, opened 2018-01-12
Python wrapping

LucasGandel:


LucasGandel:
@SimonRit 
Most of the classes have been wrapped for python. Some others should wait for the forbild branch to be integrated (because templates get modified on this branch).
I also added a very simple test to check that the wrapping process is OK. We should improve wrapping testing in the future.

For now, many warnings are thrown because of "ConstPointer" issues. This does not prevent the wrapping process to succeed and it is now fixed in ITK master.

A few ".notwrap" files have been added. Those were files generating issues that are now disabled. To fix them, we should require dimension 1 to be wrapped within ITK. We'll leave it like this for now and wait to see if more files requiring 1D come.

Issue #146, opened 2018-01-25
unknown extension for .c file


codeling:
Not sure why, but since upgrading to latest master (which made compliation with CUDA 9.1 work, thanks for that 👍 ), I get an error in CMake when configuring our project, which is referencing RTK:
```
CMake Error at <somepath>/rtk/src-git/cmake/FindCUDA_wrap.cmake:94 (try_run):
  Unknown extension ".c" for file

    <somepath>/rtk/src-git/cmake/has_cuda_gpu.c

  try_compile() works only for enabled languages.  Currently these are:

    CXX RC

  See project() command to enable other languages.
```

If I add C language to our Project such as `PROJECT (OurProject Languages CXX C)` then it works, but that wasn't required in previous versions of RTK as far as I can tell.

Anybody an idea why this has changed? Adding in C language seems to have other side effects as well so if possible I'd like to avoid it.

SimonRit:
Thanks for the report. This is caused by commit 90b25ca4 I believe but I can't reproduce the error. Which version of cmake are you using?
Maybe you can simply move this file to cxx.

codeling:
I use CMake version 3.10.2.
The suggested fix indeed works. I'm using Visual Studio 2017 15.4.5 (not the latest, 15.5.x, which doesn't work with cuda anymore), maybe that specific version (or in general building for Visual Studio) is what makes a difference?

The "other side effects" I mentioned above seem to come from somewhere else, but I have submitted a pull request for the .c -> .cxx change anyway: https://github.com/SimonRit/RTK/pull/148

PR #148, opened 2018-01-25
FIX: don't require C in dependent program CMake run

codeling:
In some configurations (e.g. when building with CMake 3.10.2 for Visual Studio 2017 15.4.5, see https://github.com/SimonRit/RTK/issues/146), when using RTK, and the dependent program doesn't have C language specified, CMake doesn't know how to handle .c files, yet would require to try_compile has_cuda.c. With this fix required programs are not required to add C to their languages.


SimonRit:
Thanks!

PR #159, opened 2018-03-31
COMP: Prepare for C++11 features in ITKv5

hjmjohnson:
ITKv5 is going to require C++11 so require newer
versions of cmake features when the CMAKE_CXX_STANDARD
flag is present so that these standards are respected.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,9 +1,14 @@
 #=========================================================
 # RTK : Reconstruction Toolkit
 #=========================================================
-cmake_minimum_required(VERSION 2.8.4 FATAL_ERROR)
-if(WIN32 OR NOT EXISTS /dev/urandom)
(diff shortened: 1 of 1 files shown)


SimonRit:
@LucasGandel ITK v5 will require cmake 3.9.5, should we align with this and skip other options?

hjmjohnson:
@SimonRit @LucasGandel 
FYI:  We (ITK developers) had a long discussion about the pro's/cons of moving to cmake 3.9.5

In the end the summary was that now was the time to move aggressively forward a large step for cmake to ensure future compatibililty with the C++11 language without a lot of extra work-arounds.

LucasGandel:
@hjmjohnson Thank you for the information.
@SimonRit I think we can definitely move forward to cmake 3.9.5.

hjmjohnson:
#161 supersedes this patch.  Thanks @LucasGandel .

PR #166, opened 2018-04-16
Move RTK binaries into RTK binary directory

LucasGandel:
When building RTK against an ITK build directory, RTK binaries
go into the RTK binary directory.
When building RTK within ITK, RTK binaries are located in the ITK
binary directory.


SimonRit:
Thanks!

PR #170, opened 2018-04-28
COMP: Fix python wrapping

LucasGandel:
Remove files using covariant vector types for now.
Clean some missing types and types redefinition in preparation for
future commits on the ITK side.
The current state should build against an ITK build tree with the following
options :
  - ITK_WRAP_double = ON;
  - ITK_WRAP_IMAGE_DIMS = 1;2;3;4;
  - ITK_WRAP_unsigned_short= ON;
  - ITK_WRAP_unsigned_char=ON
  - ITK_WRAP_vector_double=ON

WARNING: Those options needs to be set on the RTK side too when building
externally (future ITK commits should improve this).


SimonRit:
Thanks @LucasGandel!

PR #176, opened 2018-05-09
WIP: Add python wrapping missing types

LucasGandel:
Depending on the value that was set for ITK_WRAP_* options, we could have
missing types that were not required by ITK but that are necessary for building RTK.
This commit ensures that RTK-required types are always defined.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -120,27 +120,6 @@ if(NOT ITK_SOURCE_DIR)
 endif()
 
 # --------------------------------------------------------
-# Check Image Dims required for Python wrapping
-set(WRAPPING_CONFIG_ERROR_LOG "
-  Python wrapping configuration is incorrect.
-
-  Please rebuild both ITK and RTK with the following options:
-    -ITK_WRAP_IMAGE_DIMS=\"2;3;4\"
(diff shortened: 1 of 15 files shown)


LucasGandel:
@SimonRit Could you please have a close look at commit [44df753](https://github.com/SimonRit/RTK/pull/176/commits/44df753f8fcd32113cf880adba5e926f3ad74453) ? It is removing signed/unsigned comparison warnings, but I couldn't keep the 'unsigned int' type as it would cause issues when wrapping for python. Using 'int' should not change the filters behavior but I wanted to check with you.

LucasGandel:
@SimonRit I also added a first version of the 'setup.py' script to generate python packages. The description for the package in this file is still based on the ITK one. Could you please comment commit [4acb050](https://github.com/SimonRit/RTK/pull/176/commits/4acb050485549d8fddf0671b6badbf350696a273) to point me to some good description, email resources and so on ?

SimonRit:
@LucasGandel
44df753 is not a problem (although the passed values should never be negative)
4acb050 I would suggest rtk-users@openrtk.org as a contact point. For the rest, I think it's fine like this? If you want a small paragraph on RTK, you can copy paste it from the website.

PR #183, opened 2018-05-23
Sparse system matrix

tbaudier:


tbaudier:
The error on AppVeyor is:
```
Build execution time has reached the maximum allowed time for your plan (60 minutes).
```
But I do not know how to reduce compilation time

PR #186, opened 2018-06-06
Add SHA512 content link files for ExternalData storing

LucasGandel:
Testing Data have been moved to the [RTK Collection](https://data.kitware.com/#collection/5a7706878d777f0649e04776) on data.kitware.com and now use SHA512 instead of MD5.


LucasGandel:
@SimonRit We have moved testing data from midas3.kitware.com to data.kitware.com.
If you want to request an account on data.kitware.com, I can then add you as the owner of the collection in case you need to manage the collection.

SimonRit:
@LucasGandel Thanks. I'm now registered as srit on data.kitware.com.

PR #187, opened 2018-06-06
COMP: compatibility with ITK_FUTURE_LEGACY_REMOVE

dzenanz:


SimonRit:
Thanks! Is this backward compatible with ITK 4.12? This is currently the minimal ITK release for RTK and I don't think I want to break compatibility with at least 4.13 given the current instability of ITK 5.

dzenanz:
I was compiling it against 4.13. And I think the GetModifiable macro existed for years, so it should be good for 4.12.

SimonRit:
Perfect, thanks!

dzenanz:
I now turned on CUDA, and added corresponding fixes.

PR #194, opened 2018-06-26
COMP: fixed CUDA wrapping

SimonRit:
@LucasGandel: check if you agree please


LucasGandel:
@SimonRit Looks good! Thanks for fixing the wrapping!

PR #195, opened 2018-06-26
Fix testing when building as an ITK remote module

LucasGandel:


LucasGandel:
@SimonRit  This is now ready to be merged! Thanks for waiting!

PR #196, opened 2018-06-28
ENH: added examples for FirstReconstruction

SimonRit:
@LucasGandel: There are two problems on this branch
- RTK_USE_CUDA is no longer available in cmake (as discussed)
- the FirstCudaReconstruction.py does not work but I'm testing something, I'll keep you posted.


LucasGandel:
@SimonRit Sounds good! I'll have a look to fix RTKConfig.cmake

SimonRit:
@LucasGandel I have added the missing wrapping of rtk::CudaFDKConeBeamReconstructionFilter. However, the CUDA python example does not work because non GPU filters have not been wrapped for CudaImage types. Two solutions: wrapping all filters with both itk::Image and itk::CudaImage or provide a mechanism to convert an itk::Image to a itk::CudaImage. What do you think?

LucasGandel:
@SimonRit I would go for the second approach in a first time as it should be relatively simple to implement. That would give us time to make the whole process more stable.
Later we can implement the first solution as it seems cleaner and would avoid calls like
`filter->SetInput(CudaImage->GetImage())`

SimonRit:
@LucasGandel You actually don't need to do calls like this when going from CudaImage to Image since a CudaImage is an Image. It's when you go from Image to CudaImage that you need to do a conversion. I'm trying to correct the Graft method to allow grafting an Image to a CudaImage...

SimonRit:
@LucasGandel I have modified the FirstCudaExample.py using a Graft function. This now gives the same result as FirstCudaExample.cxx. This allows to use wrapped CUDA filters. Maybe we should implet SetInput(Image*) functions in CudaImageToImage filters using the same Graft mechanism? Then we would not have to do it manually.

PR #197, opened 2018-06-29
Fix missing libraries for external projects linking against RTK

LucasGandel:


LucasGandel:
@SimonRit This should allow calling find_package(RTK) to access variable such ${RTK_USE_CUDA}
I am still working on having CUDA libraries linking properly. I'll keep you posted!

SimonRit:
@LucasGandel That does not seem to work. RTK_USE_CUDA is still not known in the FirstReconstruction example... BTW, I don't have a linking problem if I remove the condition on RTK_USE_CUDA in this example.

LucasGandel:
@SimonRit The last commit should fix the failures when finding package CUDA_wrap.
You should now have access to all variables if using find_package(RTK)

SimonRit:
@LucasGandel it's a bit disappointing because we cannot use RTK as an external module, right? Shouldn't we ask the ITK guys if there is not a better solution?

With this [CMakeLists.txt](https://github.com/SimonRit/RTK/pull/196/files#diff-6b65bd9573ececa8b4ca4f72e5ae3628), I still get the following error messages

CMake Error at /home/srit/src/rtk/lin64/RTKConfig.cmake:36 (set_property):
  set_property could not find TARGET ITKCudaCommon.  Perhaps it has not yet
  been created.
Call Stack (most recent call first):
  CMakeLists.txt:7 (find_package)


CMake Error at /home/srit/src/rtk/lin64/RTKConfig.cmake:41 (set_property):
  set_property could not find TARGET ITKCudaCommon.  Perhaps it has not yet
  been created.
Call Stack (most recent call first):
  CMakeLists.txt:7 (find_package)


CMake Error at /home/srit/src/rtk/lin64/RTKConfig.cmake:45 (set_property):
  set_property could not find TARGET rtkcuda.  Perhaps it has not yet been
  created.
Call Stack (most recent call first):
  CMakeLists.txt:7 (find_package)


CMake Error at /home/srit/src/rtk/lin64/RTKConfig.cmake:49 (set_property):
  set_property could not find TARGET rtkcuda.  Perhaps it has not yet been
  created.
Call Stack (most recent call first):
  CMakeLists.txt:7 (find_package)


-- Configuring incomplete, errors occurred!

LucasGandel:
@SimonRit That might work if you set the RTK_DIR to ~/ITK_build/Modules/Remote/RTK. But you are right, it would be much better to export the remote module configuration in ITK, I'll check it!

Issue #201, opened 2018-07-20
Issues with using ChangeInformationImageFilter on CudaImage


Hakkk2002:
There seems to be a CPU-GPU synchronization issue if an `itk::ChangeInformationImageFilter<itk::CudaImage<...>>` is inserted before a Cuda filter. It works if the ChangeInformationImageFilter is inserted before a CPU filter.

The issue can be reproduced with the rtkSartCudaTest and the following patch:

```
File: include/rtkSARTConeBeamReconstructionFilter.hxx

@@ -18,10 +18,11 @@
 
 #ifndef rtkSARTConeBeamReconstructionFilter_hxx
 #define rtkSARTConeBeamReconstructionFilter_hxx
 
 #include "rtkSARTConeBeamReconstructionFilter.h"
+#include <itkChangeInformationImageFilter.h>
 
 #include <algorithm>
 
 namespace rtk
 {
@@ -282,10 +283,15 @@ SARTConeBeamReconstructionFilter<TVolumeImage, TProjectionImage>
 
   // Declare the image used in the main loop
   typename TVolumeImage::Pointer pimg;
   typename TVolumeImage::Pointer norm;
 
+  // test: insert itkChangeInformationImageFilter
+  itk::ChangeInformationImageFilter<VolumeType>::Pointer change = itk::ChangeInformationImageFilter<VolumeType>::New();
+  change->SetInput(this->GetInput(0));
+  m_ForwardProjectionFilter->SetInput(1, change->GetOutput());
+
   // For each iteration, go over each projection
   for(unsigned int iter = 0; iter < m_NumberOfIterations; iter++)
     {
     unsigned int projectionsProcessedInSubset = 0;
     for(unsigned int i = 0; i < nProj; i++)
@@ -326,11 +332,11 @@ SARTConeBeamReconstructionFilter<TVolumeImage, TProjectionImage>
         else
           pimg = m_AddFilter->GetOutput();
         pimg->Update();
         pimg->DisconnectPipeline();
 
-        m_ForwardProjectionFilter->SetInput(1, pimg );
+        change->SetInput(pimg);
         m_AddFilter->SetInput2(pimg);
         m_BackProjectionFilter->SetInput(0, m_ConstantVolumeSource->GetOutput());
         m_BackProjectionNormalizationFilter->SetInput(0, m_ConstantVolumeSource->GetOutput());
 
         projectionsProcessedInSubset = 0;
```
Results from the rtkSartCudaTest:

```
[...]

PR #202, opened 2018-07-25
BUG: Missing SetCPUBufferPointer in CudaImage<...>::SetPixelContainer(...)

Hakkk2002:
In `CudaImage<...>::SetPixelContainer(...)`, after setting a new pixel container, the CPU buffer pointer of the CudaImage's CudaDataManager shall also be updated.

 This is likely a fix for #201.


SimonRit:
Thanks (and sorry for the delay in merging)

PR #203, opened 2018-08-15
Add checks to make sure filename list is not longer than geometry

agravgaard:
In connection with [the issue on the mailing-list](https://public.kitware.com/pipermail/rtk-users/2018-August/010622.html) I have added a check in `rtkGgoFunctions.h` to make sure the vector of filenames returned by regex is not longer than the number of entries in the geometry.
The check will only occur in when `TArgsInfo` has `geometry_arg`.

I have also made sure the `VarianProBeamGeometryReader` gets the number of projections from the `Scan.xml` and checks that against the filenumber when producing geometry entries.

The check is simple, if filenames too long: remove last file from list until equivalence is reached.

However, when the check is done, I could not find a better way to get geometry entries than by reading reading it from the file. This might be unnecessary as geometry is often read later. So it could be more optimal to change the IO causality all places `SetProjectionsReaderFromGgo` is called.

If this issue from the mailing list could NOT occur for other filetypes than Xim, an alternative solution could be to make `CanReadFile` of `XimImageIO.cxx` more advanced and do an attempt at opening the file and checking for image data?

### DIFF
--- a/include/rtkGgoFunctions.h
+++ b/include/rtkGgoFunctions.h
@@ -170,7 +170,7 @@ template<typename T> struct HasGeometryArg {
 * If larger, remove the last file from the container until equal */
 template<class TArgsInfo>
 void
-CheckGeomAgainstFiles(std::vector<std::string> fileNames, typename TArgsInfo args_info, std::true_type)
+CheckGeomAgainstFiles(std::vector<std::string> &fileNames, typename TArgsInfo args_info, std::true_type)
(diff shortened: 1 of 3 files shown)


agravgaard:
This is a bad solution. I'm working on something better.

SimonRit:
Thanks for your efforts. I'll wait for your other solution if you think it can be improved.

PR #204, opened 2018-08-20
ENH: Export RTK cmake variables on find_package(ITK COMPONENTS RTK)

LucasGandel:
Set RTK_EXPORT_CODE_BUILD to specify cmake variable that should be set when
trying to find package RTK.
This allows for calling only find_package(ITK COMPONENTS RTK) when an
external builds against RTK.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -192,6 +192,34 @@ set(RTK_SYSTEM_LIBRARY_DIRS ${CMAKE_LIBRARY_OUTPUT_DIRECTORY})
 set(RTK_SYSTEM_INCLUDE_DIRS ${RTK_INCLUDE_DIRS})
 
 #=========================================================
+# Generate RTKConfig.cmake for the build tree.
+set(RTK_MODULE_PATH_CONFIG ${CMAKE_MODULE_PATH})
+set(RTK_USE_FILE "${RTK_SOURCE_DIR}/cmake/UseRTK.cmake")
+set(RTK_LIBRARY_DIRS ${CMAKE_LIBRARY_OUTPUT_DIRECTORY})
+configure_file(cmake/RTKConfig.cmake.in RTKConfig.cmake @ONLY)
+configure_file(cmake/RTKConfigVersion.cmake.in RTKConfigVersion.cmake @ONLY)
+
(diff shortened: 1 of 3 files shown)


LucasGandel:
@SimonRit This change should allow external application to build against RTK using find_package(ITK COMPONENTS RTK). Then you just have to specify the ITK_DIR used to build RTK.

For now, only RTK_USE_CUDA is specified in the export code. Please let me know if you think of any other variable that should be set when calling find_package.

SimonRit:
@LucasGandel Thanks. I'm not sure I understand. If we keep [RTKConfig.cmake.in](https://github.com/SimonRit/RTK/blob/master/cmake/RTKConfig.cmake.in), why do you need to set RTK_EXPORT_CODE_BUILD since RTK_USE_CUDA is already set in RTKConfig.cmake? If you just forgot to remove RTKConfig.cmake.in, I would say everything meaningful in it (e.g., version numbers) should go in RTK_EXPORT_CODE_BUILD.

LucasGandel:
@SimonRit Keeping RTKConfig.cmake allows to use find_package(RTK). This way is not ideal as it will ask for an ITK_DIR anyway.

Using the RTK_EXPORT_CODE variable, we can pass all the meaningful variables when calling just find_package(ITK COMPONENTS RTK). This way we just need to specify the ITK_DIR, no RTK_DIR anymore.

Removing RTKConfig.cmake will remove support for find_package(RTK), there might be some application impacted but you tell me what we should do. In the meantime I'll add the missing variable in RTK_EXPORT_CODE.

SimonRit:
@LucasGandel If I understand correctly, you have duplicated code to use find_package with ITK or RTK. I see two options:
- disable find_package(RTK). That would be a backward incompatible change but maybe not a huge problem (easy to fix and clarify the RTK change to an external module).
- we write down the content of RTK_EXPORT_CODE_BUILD to RTKConfig.cmake instead of duplicating the code.
What do you think of this second option? I also did not quite get the ITK_DIR issue.

LucasGandel:
@SimonRit You are right. I changed the content of RTKConfig.cmake.in to avoid dupplication of the code.
The second option is the way to go. We keep backward compatibility this way.

PR #205, opened 2018-08-23
Using CanReadFile to remove unreadable files

agravgaard:
This PR adds a check after the file regex function to check if all matches are readable (And some improvements to the Xim reader).

When attempting `CreateImageIO` for a file, `CanReadFile` will be called on the filename for all registered readers.
If no reader where found it returns a NULL pointer, which we can use to identify and eliminate unreadable files from the regex matches.
`CanReadFile` is supposed to fail fast, by checking file extension first. Only a files matching the extension will run the rest of the function, therefore this should be an efficient test.

In connection with [the issue on the mailing list](https://public.kitware.com/pipermail/rtk-users/2018-August/010622.html), regarding a Xim file potentially not containing image data, a check for this was added to `XimImageIO::CanReadFile` by checking the size of the image in the header.

Additionally:
I did some optimization in `XimImageIO::Read`, I have not measured the improvement, but theoretically it could give fewer cache misses, and help the compiler a bit (investigation using [godbolt.org](https://gcc.godbolt.org/z/5HaXzc) gave approx half the lines of assembly, when comparing looping though a only a single look-up-table index). The same optimizations could be applied to `HndImageIO::Read`.

### DIFF
--- a/include/rtkGgoFunctions.h
+++ b/include/rtkGgoFunctions.h
@@ -24,6 +24,7 @@
 
 #include "rtkMacro.h"
 #include "rtkConstantImageSource.h"
+#include "rtkIOFactories.h"
 #include "rtkProjectionsReader.h"
 #include <itkRegularExpressionSeriesFileNames.h>
 #include <itksys/RegularExpression.hxx>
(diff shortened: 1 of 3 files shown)


SimonRit:
Thanks for your efforts. There seems to be a test failing now (see  rtkVarianTest on travisci and circleci), can you reproduce the problem?

agravgaard:
I have managed to reproduce with GCC 8.2 on Linux. I'll try to debug :)

agravgaard:
I think I found the problem: `itk::ImageIOBase::LONG` did not behave as I expected.
Because byte size can be different on different systems.
Changing back to `UINT` seems like the best option. Negative values should be impossible anyway.

#WorksOnMyMachine with ICC 18.0 on Windows 8.1 and GCC 8.2 on Linux 4.19

SimonRit:
Thanks!

Issue #207, opened 2018-09-05
Build with ITK - install package


dkuegler:
If one selects "ITK_DIR" to be NOT the BUILD Dir of ITK, but the install dir, The error will be:

CMake Error at CMakeLists.txt:199 (include):
  include could not find load file:

    ITKModuleExternal

Configuring incomplete, errors occurred!
See also "G:/build-msvc15_x64/lib/rtk/CMakeFiles/CMakeOutput.log".
See also "G:/build-msvc15_x64/lib/rtk/CMakeFiles/CMakeError.log".

dkuegler:
This error is in the configure step of cmake

SimonRit:
RTK has moved to being an external module of ITK, maybe one cannot compile from an installed ITK. What do you think @LucasGandel?

LucasGandel:
Exactly, even if you could provide the ITKModuleExternal.cmake script from the sources, you would hit the following error message : 

> CMake/ITKModuleExternal.cmake:11:
> message(FATAL_ERROR
> "Modules can only be built against an ITK build tree; they cannot be built against an ITK install tree.")

SimonRit:
Thanks for the confirmation. I guess we need to clarify this in RTK's documentation. We are preparing a release for the coming days to be able to submit a file to ITK for compilation of RTK from within ITK as a remote module. For now, if you just clone RTK in ITK_SRC/Modules/Remote or in ITK_SRC/Modules/External, you will get a Module_RTK option that you can activate to compile RTK.
I'll leave this issue open until we have changed the error message for something more meaningful than what you currently have.

dkuegler:
This should really be a new issue, but since it is easier to explain:
I moved RTK to Modules/External and activated Module_RTK.
On Windows, when compiling ITK_CudaComman as BUILD_SHARED, I get:
C2941: definition of dllimport static data member not allowed in itkCudaContextManager.cxx
This is because ITK_CudaCommon is not defined, since the target in not defined as a shared library and therefore in the itkCudaWin32Header.h, the library is defined as __declspec(dllimport) which is bogus.

Solution: in utilities/ITKCudaCommon/CMakeLists.txt:11
add_library(ITKCudaCommon ${ITK_LIBRARY_BUILD_TYPE} ${ITKCudaCommon_SRC})

If itkcudacommon should always be static, then the declspec declarations should be removed....

dkuegler:
Also, I think the whole install infrastructure is sadly not properly set up.
[...]

PR #208, opened 2018-09-10
Shared libs with cuda

LucasGandel:


LucasGandel:
@SimonRit  This should help with issue #207; At least it helps when building RTK outside of ITK.

SimonRit:
Thanks!

PR #211, opened 2018-09-21
COMP: Fix shared libs configuration for remote

LucasGandel:
When building RTK as an ITK Remote module, the itk-module-init.cmake file
is included prior to setting the RTK_BUILD_SHARED_LIBS variable. Thus
rtkConfiguration.h was missing RTK_BUILD_SHARED_LIBS value.

Solution: Moving configuration of rtkConfiguration.h.in to ensure that all
variable are set before configuring.


LucasGandel:
@SimonRit This should be enough to fix the dashboard for Remote builds using shared libraries

PR #212, opened 2018-10-01
Allow RTK compilation without Cuda, even if Cuda is present

tbaudier:
In [nvcc-check.cmake - line 90](https://github.com/SimonRit/RTK/blob/master/cmake/nvcc-check.cmake#L90), it found Cuda and looked for the correct gcc version. If none correct gcc version were found, it raised an error. And we cannot install RTK at all with ccmake (we can using command line).

We could want to install RTK without Cuda in a such case. So I suggest to modify the fatal error:
```c++
message(FATAL_ERROR "nvcc-check: Please install adequate gcc for cuda.\nNote that gcc-4.x can be installed side-by-side with your current version of gcc.\n")
```

by a warning and set the CUDA_FOUND to OFF:
```c++
message(WARNING "nvcc-check: Please install adequate gcc for cuda.\nNote that gcc-4.x can be installed side-by-side with your current version of gcc.\n")
set(CUDA_FOUND OFF)
```


SimonRit:
Looks good to me, thanks

PR #215, opened 2018-10-08
COMP: Fix testing when building RTK as a remote module

LucasGandel:
When building RTK as a remote module, test applications should register IO
factories.


LucasGandel:
@SimonRit This should fix failing tests on the dashboard when building as remote module

PR #218, opened 2018-10-17
COMP: Change header include to robust pathing

hjmjohnson:
The vnl_include.h file should be included as vnl/vnl_include.h
to provide maximum compatibility with include directories.  This
is the path used internally in vnl as well.  By using this path
backwards compatibility is maintained, and the code is also
compatible with future versions of vnl where extraneous
include paths are now not exported.


SimonRit:
Thanks!

PR #220, opened 2018-10-31
Avoid absolute path for install targets

agravgaard:
I get an error when packaging a project with RTK as a submodule of ITK, with ITK added via. `add_subdirectory`.
My project is packaged with WiX on Windows. (I haven't tested packaging on other OS's nor packagers)
This fixes the problem for me. I guess one could give an empty `CMAKE_INSTALL_PREFIX` at configuration, to fix it as well.
However, I don't think this fix changes/breaks the ´install´ step, where `CMAKE_INSTALL_PREFIX` should be automatically added by `install()`, when given a relative path.


SimonRit:
Thanks (and sorry for the delay)

PR #224, opened 2018-11-26
Fix shared library symbol markup for Linux / Unix

thewtex:


SimonRit:
This is great, thanks a lot for looking into this issue. It works for me!

PR #225, opened 2018-11-30
COMP: fix cmake shared libs config to avoid warnings

SimonRit:
This fixes the warning:
/usr/lib64/gcc/x86_64-suse-linux/4.8/../../../../x86_64-suse-linux/bin/ld:
Warning: type of symbol `randomseed' changed from 1 to 2 in
../lib/liblpsolve55.a(myblas.c.o)

@LucasGandel Can you have a look and let me know if you (dis)agree? Thanks in advance.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -95,11 +95,10 @@ endif()
 
 # --------------------------------------------------------
 # Shared libraries option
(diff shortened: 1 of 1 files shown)


LucasGandel:
@SimonRit It should be OK but I did not test every configuration.  +1

PR #228, opened 2018-12-03
BUG: Force wrapping of ExtractPhaseImageFilter

LucasGandel:
When using itk_wrap_image_filter with dimension restriction, the
macro verifies that the dimension parameter is included in CMake
ITK_WRAP_IMAGE_DIMS variable.
Because RTK handle wrapping of dimension 1 regardless of
ITK_WRAP_IMAGE_DIMS value, this filter should not use the macro if we
want to ensure that it is wrapped for Python.


LucasGandel:
@arobert01 @SimonRit 
rtk::ExtractPhaseImageFilter should now be fixed. It was using itk_wrap_image_filter macro with a restriction on the dimension (1).
But the macro will not wrap anything if 1 is not in the ITK_WRAP_IMAGE_DIMS variable.

For the record, it was working before because we generated the packages with dimension 1 enabled in ITK.

I just checked that we can instantiate rtk.ExtractPhaseImageFilter in python but did not test its use. Please let me know if it needs further fixes.

SimonRit:
Thanks @LucasGandel. Indeed, it is all clear now.

PR #229, opened 2018-12-05
BUG: Force wrapping of type double in ExtractPhaseImageFilter

arobert01:


arobert01:
Thanks @LucasGandel !

PR #234, opened 2019-01-15
Fixes for spectral one step reconstruction

paRodesch:


kalvdans:
> BUG: remove inline CUDA function which was wrongly optimized by compiler

Interesting. Have you reported the bug to gcc?

PR #235, opened 2019-01-23
BUG: Wrong integer size for number of pixels in CUDA image and FDK kernel.

Hakkk2002:


src/rtkCudaFDKBackProjectionImageFilter.cu
```diff
@@ -72,7 +72,7 @@ void kernel_fdk(float *dev_vol_in, float *dev_vol_out, unsigned int Blocks_Y)
     }
 
   // Index row major into the volume
-  long int vol_idx = i + (j + k*c_vol_size.y)*(c_vol_size.x);
+  unsigned long long vol_idx = i + (j + (unsigned long long)k*c_vol_size.y)*(c_vol_size.x);
```
SimonRit: Thanks for the PR. I'm concerned there is an overhead changing the type in the kernel. Have you done any performance evaluation. If there is an overhead, we can template the kernel and depending on the image size, chose the index type.

utilities/ITKCudaCommon/include/itkCudaImage.hxx
```diff
@@ -45,7 +45,7 @@ void CudaImage< TPixel, VImageDimension >::Allocate(bool initializePixels)
 
   // allocate Cuda memory
   this->ComputeOffsetTable();
-  unsigned long numPixel = this->GetOffsetTable()[VImageDimension];
+  SizeValueType numPixel = this->GetOffsetTable()[VImageDimension];
```
SimonRit, line 48: I did not manage to find where ::itk::SizeValueType is defined... Any idea?

Hakkk2002: In itkIntTypes.h.
Depending on flag `ITK_USE_64BITS_IDS`, it can be either `uint64_t` or `unsigned long`

src/rtkCudaFDKBackProjectionImageFilter.cu
Hakkk2002: Maybe typedef according to ITK_USE_64BITS_IDS flag too?

SimonRit: Including itkIntTypes.h and using ::itk::SizeValueType doesn't work? Any comment on the overhead?

Hakkk2002: I've rewritten the change as
`  itk::SizeValueType i = __umul24(blockIdx.x, blockDim.x) + threadIdx.x;`
`  itk::SizeValueType j = __umul24(blockIdx_y, blockDim.y) + threadIdx.y;`
`  itk::SizeValueType k = __umul24(blockIdx_z, blockDim.z) + threadIdx.z;`
and
`  itk::SizeValueType vol_idx = i + (j + k*c_vol_size.y)*(c_vol_size.x);`

And then did a quick test. Each execution of CudaFDKBackProjectionImageFilter took on average 4.7 ms longer in the case of ITK_USE_64BITS_IDS=ON (so with type change).

SimonRit: Sounds good, thanks for testing. I'm happy to merge when I get this version. Can you also propagate the change to rtkCudaBackprojectionImageFilter.cu please?

PR #239, opened 2019-03-22
Performance: Xim and Hnd readers

agravgaard:
Until now fread was called for every compressed pixel (or pixel differences to be exact) in the file.
fread calls was the bottleneck of these readers (by intuition and according to callgrind).
Therefore reducing all these separate fread calls to one call reading the entire compressed buffer, should be more efficient, if the post-processing (decompression) of this buffer into separate pixels can be computed efficiently.

It took me a while to wrap my head around the binary operations, but I think this works.
The solution ended up being more simple than first expected, which is good because this may mean it will work on both big and little endian systems.

The reading time in "rtkvariantest" (of the respective reader->Update() calls) went from around **35-45ms|60 ms** to about **15ms|38ms (Hnd)** and **10ms|31ms (Xim)** on Linux-5.0.1 with gcc-8 in **Release|Debug** mode.

I have not tried it on windows yet, but I expect the gain may be even larger for NTFS.

### DIFF
--- a/include/rtkXimImageIO.h
+++ b/include/rtkXimImageIO.h
@@ -122,7 +122,6 @@ public:
 
 private:
   template<typename T> size_t SetPropertyValue(char *property_name, Int4 value_length, FILE *fp, Xim_header *xim);
-  template<typename T> T get_diff(char vsub, FILE* &fp);
 
   int          m_ImageDataStart;
   Int4 m_BytesPerPixel;
--- a/src/rtkHndImageIO.cxx
+++ b/src/rtkHndImageIO.cxx
@@ -18,6 +18,7 @@
 
 // std include
 #include <stdio.h>
+#include <valarray>
 
 #include "rtkHndImageIO.h"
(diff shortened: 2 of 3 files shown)


SimonRit:
Thanks, good work!

SimonRit:
Hi,
I didn't notice the 
```
#pragma omp parallel
```
but they give warnings, see the [dashboard](https://my.cdash.org/viewBuildError.php?type=1&buildid=1626021). Are they required for performances? I would trust the standard library for a reduction, e.g., `std::accumulate`. What do you think?
Thanks,
Simon

agravgaard:
You're right, std::accumulate should be preferred. I have created another pull request: https://github.com/SimonRit/RTK/pull/241

I was confused about the return type of `std::valarray<>::apply()`, which wasn't accepted by `std::begin()` and `std::end()` and gave up on making it work a little to early. 
And `std::valarray<>::sum()` would overflow for `unsigned char`

PR #241, opened 2019-03-25
Fix warnings from Hnd/Xim readers

agravgaard:
As correctly pointed out by @SimonRit in: https://github.com/SimonRit/RTK/pull/239#issuecomment-476072191
The OpenMP pragma may give warnings, and std::accumulate should be preffered.
There is no noticeable performance decrease from this change (Neither in Debug or Release on Linux 5.0.1 with gcc-8).

Also lookUpTableSize is now casted to size_t to avoid comparison of signed and unsigned.

### DIFF
--- a/src/rtkHndImageIO.cxx
+++ b/src/rtkHndImageIO.cxx
@@ -19,6 +19,7 @@
 // std include
 #include <stdio.h>
 #include <valarray>
+#include <numeric>
 
 #include "rtkHndImageIO.h"
 #include <itkMetaDataObject.h>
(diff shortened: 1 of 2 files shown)


SimonRit:
There seems to be one issue with MacOS/clang, I'll check if I can reproduce this

agravgaard:
I have tried to reproduce, but have not been able to with neither clang nor gcc on linux.
I might try again on MacOS high sierra (my desktop at home) later this week.

I suspect that apple-clang/llvm has different behavior for `std::valarray` (return type of `std::valarray<>::apply()` is implementation defined, with restrictions: https://en.cppreference.com/w/cpp/numeric/valarray/apply ). So maybe `std::vector` is the more safe option. But I'll have to check to know.

SimonRit:
I could not reproduce the bug. I relaunched the build and it passed... So let's see in the future if the bug reappears.

PR #242, opened 2019-03-27
Apply ITKv5 migration scripts

LucasGandel:


LucasGandel:
@SimonRit Opening this for discussion and for every commit to be tested each time we apply a new migration script.
Please do not merge until the WIP status is gone

SimonRit:
OK, thanks. Do you know if ITK's style has a requirement regarding alignment of type aliases? It used to have one for typedefs but it doesn't seem to be the case for type aliases...

LucasGandel:
> it doesn't seem to be the case for type aliases...

Right, looking at ITK files, it does not seem to be the case, but I cannot find an official document saying so...
IMO it is not necessary as it would not improve readability (what matters is the alias, which is always aligned after using)

LucasGandel:
@SimonRit This is ready to be reviewed. Can you please have a look?

Only one modernization change has not been applied with clang-tidy: [modernize-use-default-member-init](https://clang.llvm.org/extra/clang-tidy/checks/modernize-use-default-member-init.html)

I don't think this change would improve clarity, but please let me know if you think we should apply it.

SimonRit:
@LucasGandel Great! Thanks. If you don't have a convenient script to modernize the use of default member init, give it up. If you do, I think it's more readable and it's also probably safer to use the modern way (if you have two constructors, you don't have to duplicate the init code).

LucasGandel:
@SimonRit Sounds good! I'll do the update using https://github.com/InsightSoftwareConsortium/ITK/blob/master/Utilities/ITKv5Preparation/modernize-use-default-member-init.sh

PR #243, opened 2019-04-04
COMP: remove cmake CMP0077 warnings

SimonRit:
@LucasGandel Can you have a look to this one to remove [cmake warnings](https://my.cdash.org/viewConfigure.php?buildid=1631189)? Thanks in advance.


LucasGandel:
@SimonRit Looks good, but not sure where we are still using RTK_BUILD_SHARED_LIBS today..

SimonRit:
I would say for backward compatibility, let's merge.

Issue #238, opened 2019-02-15
Output for test in Debug/Release mode of ITKv5.0rc01 with RTK


kabcode:
Hello Simon,

I would like to give you some building and testing feedback on the ITK together with the RTK.
My system is  Windows 7 64bit, Cmake 3.13, Visual Studio 2015. TBB and GDCM were available in the latest stable version. CUDA 9.0 and VTK 8.0 are installed. 
Anything else you need to know? Do you need other files or outputs?

Here are the copied output of the test results from the Debug and RelWithDebInfo testing and the CMakeCache.
[CMakeCache.txt](https://github.com/SimonRit/RTK/files/2871089/CMakeCache.txt)
[ITK5withRTK_RelWithDebInfo.txt](https://github.com/SimonRit/RTK/files/2871082/ITK5withRTK_RelWithDebInfo.txt)
[ITK5withRTK_Debug.txt](https://github.com/SimonRit/RTK/files/2871084/ITK5withRTK_Debug.txt)

BTW thank you for your great work.

SimonRit:
Can you check #244 and tell me if that fixes your RTK issues?

SimonRit:
Can we close this issue or are you still having problems? I don't encounter problems on both linux and windows since the fix. It's just slower with is than without

kabcode:
This issue can be closed. No issues in this case anymore. Thx.

PR #244, opened 2019-04-04
FIX: use of TBB revealed bugs as pointed out in issue #238

SimonRit:


kabcode:
Hi Simon,

when I try to build the ITKv5.0rc1 the branch of the RTK is automatically switch to the tag v2.0.0.
Did you move the tag to another commit? Or bump the RTK version in the remote module otherwise?

SimonRit:
Hi,
Yes, you have to modify the RTK.remote.cmake file. Replace v2.0.0 by master or a specific tag you would like to build against. I don't think there is a better solution. I'll try to release a new RTK version soon and then I'll submit a patch to ITK.
Note that I have added a [fix](https://github.com/SimonRit/RTK/commit/4412d40025761cd5be4e34385143c11d11394206) to this PR.
Thanks,
Simon

kabcode:
Hi,

i run the tests with the build descripted in #238 in RelWithDebInfo mode and attached the results here:
[ITK5withRTKmaster_RelWithDebInfo.txt](https://github.com/SimonRit/RTK/files/3065451/ITK5withRTKmaster_RelWithDebInfo.txt)
The RTK branch was changed as you wrote in the last post. 

Still the same nine tests are failing. Unfortunately, i am not very firm how to examine the tests in detail but if there is any output information that i can get for you i'll try.

SimonRit:
Hi,
Thanks for testing. Weird... I don't know how you do this from MSVC but can you try to run the test with more log. From the command line, you can run `ctest -R "(rtk|RTK)" -VV`in your build directory.
I'll try to compile on Windows, I did the tests on Linux because my Windows box is a bit slow.
Simon

PR #245, opened 2019-04-05
STYLE: Use default member initialization

LucasGandel:
Converts a default constructor’s member initializers into the new
default member initializers in C++11. Other member initializers that match the
default member initializer are removed. This can reduce repeated code or allow
use of ‘= default’.


LucasGandel:
@SimonRit Adding default member init as discussed in #242

SimonRit:
Thanks. Pointers initialization to nullptr is superfluous for SmartPointer I believe but that does not hurt.

Issue #240, opened 2019-03-25
rtkConfiguration.h is not installed when RTK 2.0 is build as a submodule of ITK 5.0 RC 1


Siron777:
I have built ITK 5.0 RC 1 on visual studio 2017 15.9.7 with RTK 2.0 as a submodule. When I build a project against the installation of this build, it does not find the header rtkConfiguration.h which is included in rtkConstantImageSource.h. The headers are present in the build folder of ITK, but not in the installation folder.

Version
5.0rc1, with RTK 2.0

Environment

Windows 10
Visual studio 2017 15.9.7
Compilation in 64 bits
Ninja
cmake command: cmake "D:/dev/library/source/ITK-5.0rc01" -G "Ninja" -DCMAKE_INSTALL_PREFIX:PATH=D:/dev/library/x86-64_msvc2017/itk/5.0_rc1 -DModule_RTK=ON -DRTK_BUILD_APPLICATIONS=OFF -DCMAKE_BUILD_TYPE:STRING=RelWithDebInfo -DCMAKE_DEBUG_POSTFIX=d -DBUILD_SHARED_LIBS=OFF -DBUILD_TESTING=OFF -DCUDA_TOOLKIT_ROOT_DIR=D:/dev/conan_c/Cuda/10.0/AI/extern/package/479e5c321685f6e20cbbf93c38ae334c54ade580

SimonRit:
@LucasGandel The whole installation architecture of RTK is really messy and I have the feeling it is not done at it should. Installation variables are defined in the main CMakeLists.txt ([here](https://github.com/SimonRit/RTK/blob/master/CMakeLists.txt#L46)) but redefined in the macro itk_module_impl. As pointed out in this issue, rtkConfiguration.h is not installed and I'm not sure the rest is installed properly either. Do you know how to clean this?

LucasGandel:
@SimonRit You are right, I'll add this to the current to-do list.
Adding the missing rtkConfiguration.h file should be fairly easy. Cleaning the overall process is trickier as we want to support both standalone installation and installation within ITK.

SimonRit:
[...]

PR #246, opened 2019-04-15
BUG: Install rtkConfiguration.h

LucasGandel:


LucasGandel:
@SimonRit This should fix issue #240.
The global architecture for the installation of RTK within ITK is OK.
The only thing I noticed is that lpsolve gets installed in its own RTK directory.

In the meantime I am trying to setup a clean way to install RTK when building outside of ITK but I am not sure we can support this.

SimonRit:
Ok, it seems to work for me. I'll merge and close the issue until the next installation proble.... Thanks a lot @LucasGandel !

Issue #250, opened 2019-05-20
How to install RTK v2.0.0 as a ITK module


lucysrausch:
Hey,

I would like to get started with this promising project, however I'm a bit confused by the installation process documented in your wiki: http://wiki.openrtk.org/index.php/RTK_wiki_help

I assume the documentation is outdated, as it relies on providing RTK the path to the ITK binaries.
However, in the release note for v2.0.0 you state it is now possible and preferred to build RTK as a module of ITK.

Can you please provide me the correct way of setting up RTK for the latest version?

Thanks a lot,
Niklas

codeling:
Here's a very quick intro how it worked for my project:

- Get ITK 5.0.0
- In CMake configuration, enable "Module_RTK".
- In your project's CMake configuration, instead of "FindPackage(RTK)", add "RTK" to the components in the "FindPackage(ITK...)" call, e.g. "FindPackage(ITK COMPONENTS RTK)".

If you use INCLUDE(${ITK_USE_FILE}), you afterwards should be able to access all RTK functionality.

SimonRit:
Fixed by #254

PR #254, opened 2019-07-24
DOC: first version of the installation instructions

SimonRit:
This is an attempt to fix issue #250 (sorry for the delay). @arobert01 @LucasGandel @codeling @NiklasFauth @acoussat can you review and comment before I merge? This file will replace the first section of the [wiki](http://wiki.openrtk.org).


SimonRit:
Sorry, there was a problem with my commit. I don't know how I did this but it's now in the PR and the file is visible [here](https://github.com/SimonRit/RTK/blob/a72c00a4120431ee60f36db0adda3da886147357/documentation/INSTALLATION.md).

arobert01:
Looks good to me

codeling:
Haven't tried all details, but looks good to me!
The only thing I could think of that could be added is something regarding adding RTK to the COMPONENTS section of cmake's FindPackage call in an application using RTK (see https://github.com/SimonRit/RTK/issues/250#issuecomment-514516784)

LucasGandel:
Looks good to me! 
+1 for adding a section about find_package as suggested by codeling

acoussat:
Looks good to me! Nice that you list and explain the RTK-specific CMake options, I have learned something new along the way.

SimonRit:
Thanks for your feedbacks and sorry for the delay. I have added a GettingStarted.md file and I have added a link to the line containing find_package of the FirstReconstruction CMakeLists.txt. It's now merged.

PR #260, opened 2019-10-01
One step streaming

SimonRit:
@paRodesch Can you check that this change reduces memory consumption but does not affect the result? The result should be numerically identical.


paRodesch:
@SimonRit : I tested the last version on the example here :
https://www.creatis.insa-lyon.fr/~rodesch/OS_exemple/

But I observe a numerical difference between case with/without streaming

PR #265, opened 2019-11-06
wrapping for rtkMechlemOneStepSpectralReconstructionFilter with 2 or …

paRodesch:
Modification of wrapping file to enable use of rtkMechlemOneStepSpectralReconstructionFilter with 2 or 3 materials. 
_nmat_ variable is used either with 2 or 3 in foreach loop.


SimonRit:
Thanks, I tested it and it works. Can you amend your commit message and prefix it with ENH? I try to follow [ITK recommendations](https://github.com/InsightSoftwareConsortium/ITK/blob/master/CONTRIBUTING.md)

SimonRit:
Thanks!

PR #266, opened 2019-11-08
Iteration reporting

acoussat:
First steps towards an iteration reporting system in RTK. The goal is to allow users to plug their own custom callback called on each completion of any iteration filter. This is still a work in progress.

So far only [`rtkSARTConeBeamReconstructionFilter`](https://github.com/acoussat/RTK/blob/iteration-reporting/include/rtkSARTConeBeamReconstructionFilter.h) and [`rtkConjugateGradientConeBeamReconstructionFilter`](https://github.com/acoussat/RTK/blob/iteration-reporting/include/rtkConjugateGradientConeBeamReconstructionFilter.h) have been updated. The simplest way of testing the reporting is by using `rtksart` with the `-v` (verbose) flag.

Allows filters to [observe another filter](https://github.com/acoussat/RTK/blob/dc9c02076b3d818b9061ec67bfb5d5ded5d26fef/applications/rtksart/rtksart.cxx#L124) and invoke a callback upon iteration completion. [Some default callbacks](https://github.com/acoussat/RTK/blob/iteration-reporting/include/rtkIterationCommands.h) were defined but you can easily define your own callbacks by deriving [`IterationCommand`](https://github.com/acoussat/RTK/blob/iteration-reporting/include/rtkIterationCommands.h#L36).

This is done by exploiting the possibilities of [`itk::IterationReporter`](https://itk.org/Doxygen/html/classitk_1_1IterationReporter.html). Note that [`itk::ProgressReporter`](https://itk.org/Doxygen/html/classitk_1_1ProgressReporter.html) also exists but using [`itk::IterationReporter`](https://itk.org/Doxygen/html/classitk_1_1IterationReporter.html) seems more semantically correct to me.

Also note [the trick](https://github.com/acoussat/RTK/blob/iteration-reporting/include/rtkConjugateGradientConeBeamReconstructionFilter.hxx#L394) used for [`rtkConjugateGradientConeBeamReconstructionFilter`](https://github.com/acoussat/RTK/blob/iteration-reporting/include/rtkConjugateGradientConeBeamReconstructionFilter.h). The filter listens to its sub-filter `m_ConjugateGradientFilter` and automatically passes up the notification.

Passing a callback should work using the Python wrapping as well, like so:
```python3
def callback():
  pass # code of your callback
filter.AddObserver(itk.IterationEvent(), callback)
```

Things left to do:
- [ ] decide whether to use [`itk::IterationReporter`](https://itk.org/Doxygen/html/classitk_1_1IterationReporter.html) or [`itk::ProgressReporter`](https://itk.org/Doxygen/html/classitk_1_1ProgressReporter.html)
- [ ] implement the notification mechanism in all iterative filters

### DIFF
--- a/applications/rtkadmmwavelets/rtkadmmwavelets.cxx
+++ b/applications/rtkadmmwavelets/rtkadmmwavelets.cxx
@@ -118,6 +118,8 @@ int main(int argc, char * argv[])
 
   admmFilter->SetDisableDisplacedDetectorFilter(args_info.nodisplaced_flag);
 
+  REPORT_ITERATIONS(admmFilter, ADMM_Wavelets_FilterType, OutputImageType);
+
   TRY_AND_EXIT_ON_ITK_EXCEPTION( admmFilter->Update() )
 
   // Set writer and write the output
--- a/applications/rtkconjugategradient/rtkconjugategradient.cxx
+++ b/applications/rtkconjugategradient/rtkconjugategradient.cxx
@@ -21,6 +21,7 @@
 
 #include "rtkThreeDCircularProjectionGeometryXMLFile.h"
 #include "rtkConjugateGradientConeBeamReconstructionFilter.h"
+#include "rtkIterationCommands.h"
 
 #include <iostream>
 #include <fstream>
@@ -135,6 +136,8 @@ int main(int argc, char * argv[])
   conjugategradient->SetGeometry( geometryReader->GetOutputObject() );
   conjugategradient->SetNumberOfIterations( args_info.niterations_arg );
   conjugategradient->SetDisableDisplacedDetectorFilter(args_info.nodisplaced_flag);
+
+  REPORT_ITERATIONS(conjugategradient, ConjugateGradientFilterType, OutputImageType)
 
   TRY_AND_EXIT_ON_ITK_EXCEPTION( conjugategradient->Update() )
 
--- a/applications/rtkconjugategradient/rtkconjugategradient.ggo
+++ b/applications/rtkconjugategradient/rtkconjugategradient.ggo
@@ -14,4 +14,3 @@ option "nocudacg"       - "Do not perform conjugate gradient calculations on GPU
 option "mask"           m "Apply a support binary mask: reconstruction kept null outside the mask)"                   string no
 option "costs"          - "Show residual costs at each iteration at the end of the process"                           flag   off
 option "nodisplaced"    - "Disable the displaced detector filter"                                                     flag   off
-
--- a/include/rtkSARTConeBeamReconstructionFilter.hxx
+++ b/include/rtkSARTConeBeamReconstructionFilter.hxx
@@ -22,6 +22,7 @@
 #include "rtkSARTConeBeamReconstructionFilter.h"
 
 #include <algorithm>
+#include <itkIterationReporter.h>
 
 namespace rtk
 {
@@ -276,6 +277,8 @@ SARTConeBeamReconstructionFilter<TVolumeImage, TProjectionImage>::GenerateData()
   typename TVolumeImage::Pointer pimg;
   typename TVolumeImage::Pointer norm;
 
+  itk::IterationReporter iterationReporter(this, 0, 1); // report every iteration
+
   // For each iteration, go over each projection
   for (unsigned int iter = 0; iter < m_NumberOfIterations; iter++)
   {
@@ -341,8 +344,9 @@ SARTConeBeamReconstructionFilter<TVolumeImage, TProjectionImage>::GenerateData()
         m_BackProjectionNormalizationFilter->SetInput(0, norm);
       }
     }
+    this->GraftOutput(pimg);
+    iterationReporter.CompletedStep();
   }
-  this->GraftOutput(pimg);
 }
 
 } // end namespace rtk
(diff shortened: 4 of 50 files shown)


wrapping/itkIterationReporter.wrap
```diff
@@ -0,0 +1 @@
+itk_wrap_simple_class("itk::IterationReporter")
```
SimonRit: Should this be submitted to ITK? Is it actually usable from python?

include/rtkSARTConeBeamReconstructionFilter.hxx
```diff
@@ -348,6 +351,7 @@ SARTConeBeamReconstructionFilter<TVolumeImage, TProjectionImage>
         m_BackProjectionNormalizationFilter->SetInput(0, norm);
         }
       }
+    iterationReporter.CompletedStep();
```
SimonRit: The choice here is to report only iterations, not subsets. Should we consider reporting subsets as well?

applications/rtkconjugategradient/rtkconjugategradient.ggo
```diff
@@ -14,4 +14,5 @@ option "nocudacg"       - "Do not perform conjugate gradient calculations on GPU
 option "mask"           m "Apply a support binary mask: reconstruction kept null outside the mask)"                   string no
 option "costs"          - "Show residual costs at each iteration at the end of the process"                           flag   off
 option "nodisplaced"    - "Disable the displaced detector filter"                                                     flag   off
-
+option "output-every"   - "Output intermediate reconstruction after some iterations"                                  int    no
```
SimonRit: How come you have these options for rtkconjugategradient and not for rtksart? Maybe consider doing a section ggo ( similar to those in the applications directory).

SimonRit (changes requested):
Looks good! Just a couple of questions / comments. I guess we should propagate this to all iterative algorithms if possible?

wrapping/itkIterationReporter.wrap
acoussat: As far as I tested it, yes it is actually usable from Python. I have no idea if that line of code fits more into ITK's code than here though.

include/rtkSARTConeBeamReconstructionFilter.hxx
acoussat: It would be nice, but I don't understand how that would work from the user's side to differentiate between `IterationEvent`s being emitted by the completion of an iteration or a subset. I guess one might create another event (`SubsetEvent`?). Do you have any other idea?

applications/rtkconjugategradient/rtkconjugategradient.ggo
acoussat: That was for testing purpose. Is such a behavior desired for every iterative filter?

SimonRit: Yes but it should be off by default (which does not seem to be the case yet).

include/rtkSARTConeBeamReconstructionFilter.hxx
SimonRit: You're right and I don't have a better idea... Let's leave it like this yet and implement a solution when someone needs it.

wrapping/itkIterationReporter.wrap
SimonRit: Yes, it fits better in ITK's code. Try submitting it and we'll see how they feel about it. See [CONTRIBUTING.md](https://github.com/InsightSoftwareConsortium/ITK/blob/master/CONTRIBUTING.md).

acoussat: Done [here](https://github.com/InsightSoftwareConsortium/ITK/pull/1423), let's see what they think about it!

applications/rtkadmmwavelets/rtkadmmwavelets.cxx
```diff
@@ -28,6 +28,7 @@
 #include "rtkConstantImageSource.h"
 #include "rtkDisplacedDetectorImageFilter.h"
 #include "rtkADMMWaveletsConeBeamReconstructionFilter.h"
+#include "rtkIterationCommands.h"
```
SimonRit: I would move all `#include "rtkIterationCommands.h"` to rtkMacro.h

applications/rtkconjugategradient/rtkconjugategradient.cxx
```diff
@@ -136,6 +137,29 @@ int main(int argc, char * argv[])
   conjugategradient->SetNumberOfIterations( args_info.niterations_arg );
   conjugategradient->SetDisableDisplacedDetectorFilter(args_info.nodisplaced_flag);
 
+  if(args_info.verbose_flag)
```
SimonRit: It seems that you haven't applied the new clang format... {} are now aligned with the if.

applications/rtkconjugategradient/rtkconjugategradient.ggo
SimonRit: Still not resolved, right?

SimonRit (commented):
Thanks. It seems to be close to a merge, right? If yes, I think you can prepare a one commit PR.

applications/rtkconjugategradient/rtkconjugategradient.ggo
acoussat: Not yet, I should hopefully finish that before the end of the week! Sorry about the long time taken.

applications/rtkadmmwavelets/rtkadmmwavelets.cxx
acoussat: Done in [commit b7710957e48a620f8c526a1445c82b46f3effd45](https://github.com/SimonRit/RTK/pull/266/commits/b7710957e48a620f8c526a1445c82b46f3effd45), thanks for the tip!

applications/rtkconjugategradient/rtkconjugategradient.ggo
acoussat: [Commit a8860d5c78fb7c6cdfecf8a5547c3e64d6943287](https://github.com/SimonRit/RTK/pull/266/commits/a8860d5c78fb7c6cdfecf8a5547c3e64d6943287) implements that in all iterative filters! (I admit I haven't thoroughly tested it on every filter.)

applications/rtkconjugategradient/rtkconjugategradient.cxx
acoussat: Commits [0bc6b4d2c17a22be9f310dc81484cd075f26c504](https://github.com/SimonRit/RTK/pull/266/commits/0bc6b4d2c17a22be9f310dc81484cd075f26c504) and [7d6ad8ecd290072e3d79da46e5922bbb1a0c9bff](https://github.com/SimonRit/RTK/pull/266/commits/7d6ad8ecd290072e3d79da46e5922bbb1a0c9bff) should fix this.

acoussat:
I think everything should be alright now. Tell me if anything breaks or if the code feels weird!

acoussat:
Squashed all commits together. Next step would be to somehow document this feature and also add some tests.

PR #268, opened 2019-11-21
Adding previously available Hnc reading capabilities to RTK 2.0.1

GabrieleBelotti:
First commit which added rtkHncImageIO, rtkHncImageIOFactory and rtkVarianObiHncRawToAttenuationImageFilter.

Changes to CMakeLists of master and src have been made to include BZIP2 directories only in the event of a flag RTK_USE_BZIP2 been set.
RTK_USE_BZIP2 is defined in rtkConfiguration.


SimonRit:
Thanks. There are still some errors, see e.g. [this build](https://my.cdash.org/buildSummary.php?buildid=1732725).
I have one question: are you going to use the .hnd.bz2 extension? Because it seems to me that we should completely remove the bzip2 code, the unzipping can be done outside RTK.

GabrieleBelotti:
> Thanks. There are still some errors, see e.g. [this build](https://my.cdash.org/buildSummary.php?buildid=1732725).
> I have one question: are you going to use the .hnd.bz2 extension? Because it seems to me that we should completely remove the bzip2 code, the unzipping can be done outside RTK.

It's not in my interest to use it so I will remove it at once, maybe we'll avoid further complications

GabrieleBelotti:
> > Thanks. There are still some errors, see e.g. [this build](https://my.cdash.org/buildSummary.php?buildid=1732725).
> > I have one question: are you going to use the .hnd.bz2 extension? Because it seems to me that we should completely remove the bzip2 code, the unzipping can be done outside RTK.
> 
> It's not in my interest to use it so I will remove it at once, maybe we'll avoid further complications

Actually, let me first commit a fix to the error, and then I'll make the necessary changes.
If there are other conflicts with bzip2 we can remove it definetely. What do you think?

SimonRit:
I think we should remove the dependency to bzip2, even if it's optional. I don't see a good reason to include it in RTK.

GabrieleBelotti:
> I think we should remove the dependency to bzip2, even if it's optional. I don't see a good reason to include it in RTK.

Ok, I'll make the necessary changes and push the commit!

SimonRit:
I think a large part of the code is duplicated from Hnd reader. If you can provide an Hnc file I could try to avoid duplicating large chunks of code.

GabrieleBelotti:
> I think a large part of the code is duplicated from Hnd reader. If you can provide an Hnc file I could try to avoid duplicating large chunks of code.

https://we.tl/t-21gwOxdcmm 
Here's a projection in .hnc.
If you may need it I have a set of .hnc projections with the relative geometry file, but I'd rather share it privately. In this case just mail me and I'll send a link asap.

SimonRit:
Sorry, I failed to work on this before. I'll give up removing the duplicated code and take it as is after some clean up. But I can't use the file you sent to write a test without the corresponding `norm.hnc` file used in include/rtkVarianObiHncRawToAttenuationImageFilter.hxx. Do you also have such a file? Is it always provided along hnc files?

GabrieleBelotti:
> Sorry, I failed to work on this before. I'll give up removing the duplicated code and take it as is after some clean up. But I can't use the file you sent to write a test without the corresponding `norm.hnc` file used in include/rtkVarianObiHncRawToAttenuationImageFilter.hxx. Do you also have such a file? Is it always provided along hnc files?

Sorry this went totally over my head. I don't have a norm.hnc file since my scanner is not directly producing .hnc as an output format but rather we are converting .raw to .hnc having the template of the file. Instead of using rtkVarianObiHncRawToAttenuationImageFilter I am personally using rtkLUTbasedVariableI0RawToAttenuationImageFilter to convert raws to attenuation. I realize this is unhelpful but maybe https://github.com/ZijunH has some valid data to share

SimonRit:
So maybe I should simply remove rtkVarianObiHncRawToAttenuationImageFilter.h*?

GabrieleBelotti:
> So maybe I should simply remove rtkVarianObiHncRawToAttenuationImageFilter.h*?

Personally, I would do that, since the format seems to be deprecated I reckon it wouldn't be a problem to exclude the need for a norm chamber init file

PR #279, opened 2019-12-18
ENH: Wrap NumPy bridge for vector components 2;3;4;5

LucasGandel:
Setup infrastructure to add missing types to the itkBridgeNumPy module.

Unlike ITK, we can not use PyBuffer.i.init to define global function.
This only works when itkPyBuffer is the only submodule in the module being
wrapped, which is not the case when wrapping within RTK module.

WRAPPER_SWIG_LIBRARY_FILES is used to tell swig that additional .i files
should be included in the generated module.i file.
Unfortunately this variable can not operate at a submodule level and will
be used for all submodules when wrapping the current module.

This results in a mix of %include and %import rules which ends up not
replacing the expected chunk in the generated file.

To work around this problem, we put everything in the %extend section, that
will be only replaced for the specified class and won't affect other
submodules.

### DIFF
--- a/wrapping/CMakeLists.txt
+++ b/wrapping/CMakeLists.txt
@@ -1,4 +1,5 @@
 itk_wrap_module(RTK)
+
 set(WRAPPER_SUBMODULE_ORDER
   itkFixedArrayRTK
   itkVectorRTK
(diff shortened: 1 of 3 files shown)


SimonRit:
We will implement it as suggested by @thewtex on [discourse](https://discourse.itk.org/t/additional-python-types-for-itk-remote-external-modules/2509/6) but in the mean time, I suggest that we merge this PR. Agreed @LucasGandel ?

LucasGandel:
@SimonRit Agreed!

PR #280, opened 2019-12-19
ENH: Setup self hosted pipeline

LucasGandel:


LucasGandel:
@SimonRit 
This PR should be ready soon. Tests and packages generation [are running](https://dev.azure.com/lucasgandel/lucasgandel/_build).
Once both builds are successful (I'll let you know) you can integrate and update the pipelines on the dev.azure RTK organization:
- create a new pipeline using RTK\test\ContinuousIntegration\AzurePipelinesPythonPackages.yml
- delete the existing "Build and Test" and create a new one with RTK\test\ContinuousIntegration\AzurePipelinesBuildAndTest.yml

LucasGandel:
CI and packages generation are working fine! I think you can safely integrate this and I will improve self-hosted pipelines in another PR.

SimonRit:
Thanks a lot! I have just changed the file for the "Build and Test" instead of deleting and creating. The Python packages run manually right? I need to launch it when I need packages?
Maybe it's time to modify use Azure pipelines badges now? Should we remove other CIs?

LucasGandel:
@SimonRit Thanks for setting up the pipelines! I'll try to push a PR introducing Azure pipeline badges this week.
We can definitely remove appveyor and circleCI but I like that you test CUDA builds on Travis so I would keep it for now.
I don't see any restriction in running the Packages pipeline automatically (except if there are some Azure limits in terms of build time/size I am not aware of).
My idea was to make it automatic and add post build tasks to test and update the packages on PyPI. This might come once we have a python tests in RTK.

PR #286, opened 2020-01-15
ENH: First simple test for iteration reporting

acoussat:
Adds a simple test for images output during iteration reporting. This test checks that MSE decreases during the three first OSEM iterations.


SimonRit:
Thanks! The test seems to fail on circleci, I'm not sure why but it seems to miss one file. There is also a warning to solve, see [here](https://my.cdash.org/viewBuildError.php?type=1&buildid=1759545).
On a higher level, I'm not too fond of creating one application per application test. Can we try to have one application to check image quality? It could have options passed by ggo to set, e.g., thresholds.

acoussat:
Thanks for the comments! Could you detail how you would expect the test to work? Should I somehow merge the application `rtkcheckiterationsquality` into `rtkcheckimagequality`? As far as I understand, `rtkcheckimagequality` only checks whether a given image contains something. Should we extend its possibilities? If yes, could you provide some kind of example command line that should work? Thanks!

acoussat:
Commit `8f2b2b2` suggests some improvements, but I'm struggling to properly test iterations with my limited comprehension of CMake. More exactly, how can I test all of my iterations in a single CMake `add_test` instance? This is still unclear to me.

applications/CMakeLists.txt
```diff
@@ -108,33 +110,33 @@ if(BUILD_TESTING)
...
   endif()
 
   # Iteration reporting testing
-  add_test(rtkdrawshepploganphantomtest ${CMAKE_RUNTIME_OUTPUT_DIRECTORY}/rtkdrawshepploganphantom -o reference.mha --dimension 84 --phantomscale 40)
-
   add_test(rtkapposemtest ${CMAKE_RUNTIME_OUTPUT_DIRECTORY}/rtkosem -g geo -p . -r sheppy.mha -o osem.mha -n 3 --dimension 84 --output-every 1 --nprojpersubset 45 --iteration-file-name osem%d.mha)
   set_tests_properties(rtkapposemtest PROPERTIES DEPENDS rtkappprojectshepploganphantomtest)
 
-  add_test(rtkapposemchecktest ${CMAKE_RUNTIME_OUTPUT_DIRECTORY}/rtkcheckiterationsquality reference.mha osem1.mha osem2.mha osem3.mha)
+  add_test(rtkapposemchecktest ${CMAKE_RUNTIME_OUTPUT_DIRECTORY}/rtkcheckimagequality -i reference.mha -j osem1.mha -t 35000)# && ./rtkcheckimagequality -i reference.mha -j osem2.mha -t 16000 && ./rtkcheckimagequality -i reference.mha -j osem3.mha -t 10000)
```
acoussat: How can I test all these cases in one go? I have tried using `&&`, `;`, `test` and even some weird combination of the three, but none of them succeeded.

SimonRit: Thanks, looks much better IMO. If you feel that more than one iteration should be tested, can't you add one add_test per iteration? E.g.
```
add_test(rtkapposemchecktest1 ${CMAKE_RUNTIME_OUTPUT_DIRECTORY}/rtkcheckimagequality -i reference.mha -j osem1.mha -t 35000)
add_test(rtkapposemchecktest2 ${CMAKE_RUNTIME_OUTPUT_DIRECTORY}/rtkcheckimagequality -i reference.mha -j osem2.mha -t 16000)
add_test(rtkapposemchecktest3 ${CMAKE_RUNTIME_OUTPUT_DIRECTORY}/rtkcheckimagequality -i reference.mha -j osem3.mha -t 10000)
```
Don't hesitate to ask the question on [CMake's discourse](https://discourse.cmake.org/) if you'd rather have one test only.

SimonRit: Another suggestion: try [add_custom_command](https://cmake.org/cmake/help/latest/command/add_custom_command.html) to execute several commands and use this custom command in your test.

acoussat: > Don't hesitate to ask the question on [CMake's discourse](https://discourse.cmake.org/) if you'd rather have one test only.

[I have asked on CMake's Discourse](https://discourse.cmake.org/t/executing-a-complex-command-in-ctest/541), and [was suggested to create a custom CMake script](https://discourse.cmake.org/t/executing-a-complex-command-in-ctest/541/2). it seems a little bit overkill to me for such a trivial test. What do you feel? Should I simply create three separate tests?

SimonRit: Yes I agree. Another option: make the -i, -j and -t options a "multiple" option in gengetopt and loop over them in the cxx code. The test would then look like
```
add_test(rtkapposemchecktest ${CMAKE_RUNTIME_OUTPUT_DIRECTORY}/rtkcheckimagequality -i reference.mha,reference.mha,reference.mha -j osem1.mha,osem2.mha,osem3.mha -t 35000,16000,10000
```

acoussat: Commit f06eb1b implements this solution. Tell me what you think!

applications/rtkcheckimagequality/rtkcheckimagequality.cxx
```diff
@@ -0,0 +1,113 @@
...
+  ImageIteratorType itRef( reference, reference->GetBufferedRegion() );
+
+  using ErrorType = double;
+  ErrorType EnerError = 0.;
+
+  itTest.GoToBegin();
+  itRef.GoToBegin();
+
+  while( !itRef.IsAtEnd() )
+    {
```
SimonRit: This file seems to need a clang-format?

SimonRit (commented):
Thanks, sorry for the delay. Can you just check the clang-format of the cxx file before I merge?

SimonRit:
Thanks. Should the commits be squashed in one?

SimonRit:
Rebased onto master and pushed (b441fc7b1f79053542cb37fe2a1e04f6227b43cb).

Issue #290, opened 2020-01-23
Inconsistency in corner calculations for displaced detector


SimonRit:
Issue raised by Gabriele Belotti on the mailing list: "On a smaller note, there is some inconsistency in the inferior/superior image corner calculation between positive and negative isocenter projection displacements of the same entity (say + or - 140), which in turn affect “theta” calculation."

GabrieleBelotti:
I've found the origin of this inconsistency:
https://github.com/SimonRit/RTK/blob/45af4dd4341ba33ed216d54c6a7aeaff5d0d1fd5/include/rtkDisplacedDetectorImageFilter.hxx#L118-L120

Subtracting 1 to the size creates a different starting guess for positive and negative displacements.
From my understanding this is a small bug in the calculation of the corners of the image and the -1 should be removed from both lines (either negative and positive spacing).

Side note: I will include this quick fix in https://github.com/SimonRit/RTK/pull/291 as this solves my reconstruction problems (the artifacts is now absent)

SimonRit:
Closing because it could not be reproduced, see conversion in #291.

PR #291, opened 2020-01-27
Added DualRotation to displaced detector filters (Cuda and Cpu)

GabrieleBelotti:
Added DualRotation flag to displaced detector filters (can be set using a get/set standard method), additional effort has been made to provide a weighting consistent with the current implementation to projections stacks described by a geometry file generated by two rotations with opposite detector displacements.

The weighting is compatible with Parker Short Scan weighting filters, so it is possible to directly test results, provided a geometry file with a dual rotation (an example of suche file is attached --> 600 projections on a range of 220° limited ROM and +/-140 displacement of isocenter projection on X axis).

[geometry_hf_dual_220_300.zip](https://github.com/SimonRit/RTK/files/4117461/geometry_hf_dual_220_300.zip)

Example of PSSF output is also attached as reference:

OutputExamplewithPSSF

Additional notes : 
- Presence of an artifact in the center of the reconstructed volume which is more intense as the detector displacement increases
- Especially in the cuda implementation code can be reduced, I've tried my best to safely implement without damaging the existing implementation
- Implementation absent of OffsetFOV weighting filters

### DIFF
--- a/include/rtkCudaDisplacedDetectorImageFilter.hcu
+++ b/include/rtkCudaDisplacedDetectorImageFilter.hcu
@@ -35,4 +35,22 @@ CUDA_displaced_weight(int     proj_idx_in[3],
                       float   proj_row,
                       float   proj_col);
 
+void
+CUDA_displaced_weight_dual(int     proj_idx_in[3],
+						   int     proj_dim_in[3],
+						   int     proj_dim_in_buf[2],
+						   int     proj_idx_out[3],
+						   int     proj_dim_out[3],
+						   int     proj_dim_out_buf[2],
+						   float * dev_proj_in,
+						   float * dev_proj_out,
+						   float * geometries,
+						   float   theta,
+						   bool    isPositiveCase,
+						   float   proj_orig,
+						   float   proj_row,
+						   float   proj_col,
+						   float   InferiorCorner_2,
+						   float   SuperiorCorner_1);
+
 #endif //__rtkCudaDisplacedDetectorImageFilter_hcu
--- a/include/rtkDisplacedDetectorImageFilter.h
+++ b/include/rtkDisplacedDetectorImageFilter.h
@@ -108,6 +108,12 @@ public:
   itkGetMacro(Disable, bool);
   itkSetMacro(Disable, bool);
 
+  /** Get / Set the Dual Rotation parameter
+   */
+  itkGetMacro(DualRotation, bool);
+  itkSetMacro(DualRotation, bool);
+
+
 protected:
   DisplacedDetectorImageFilter();
 
@@ -116,6 +122,13 @@ protected:
   /** Retrieve computed inferior and superior corners */
   itkGetMacro(InferiorCorner, double);
   itkGetMacro(SuperiorCorner, double);
+
+  /** Retrieve computed inferior and superior corners - Dual Rotation positive shift */
+  itkGetMacro(InferiorCorner_1, double);
+  itkGetMacro(SuperiorCorner_1, double);
+  /** Retrieve computed inferior and superior corners - Dual Rotation negative shift*/
+  itkGetMacro(InferiorCorner_2, double);
+  itkGetMacro(SuperiorCorner_2, double);
 
   void
   GenerateInputRequestedRegion() override;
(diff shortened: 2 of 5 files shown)


SimonRit:
Thanks. Why not just split the projection stack in two and apply the displaced detector filter without modifications on each substack separately? Doesn't this work? Here is a sample code with command lines
```
head -2705 geometry_hf_dual_220_300.xml >rot1.xml
tail -1 geometry_hf_dual_220_300.xml >>rot1.xml
  
head -5 geometry_hf_dual_220_300.xml >rot2.xml
tail -2701 geometry_hf_dual_220_300.xml >>rot2.xml

rtkprojectshepploganphantom -g rot1.xml -o proj1.mha --dimension 512,3
rtkprojectshepploganphantom -g rot2.xml -o proj2.mha --dimension 512,3
rtkfdk --dimension 256,1,256 -p . -r proj1.mha -o fdk1.mha -g rot1.xml
rtkfdk --dimension 256,1,256 -p . -r proj2.mha -o fdk2.mha -g rot2.xml
clitkImageArithm -i fdk1.mha -j fdk2.mha -o fdk.mha
```
where clitkImageArithm is a tool part of [vv](http://vv.creatis.insa-lyon.fr) which simply adds the two volumes. This works perfectly without 1a20fb42 so it seems that this patch is not required. There must be something wrong in your input data?

GabrieleBelotti:
> Thanks. Why not just split the projection stack in two and apply the displaced detector filter without modifications on each substack separately? Doesn't this work? Here is a sample code with command lines
> 
> ```
> head -2705 geometry_hf_dual_220_300.xml >rot1.xml
> tail -1 geometry_hf_dual_220_300.xml >>rot1.xml
>   
> head -5 geometry_hf_dual_220_300.xml >rot2.xml
> tail -2701 geometry_hf_dual_220_300.xml >>rot2.xml
> 
> rtkprojectshepploganphantom -g rot1.xml -o proj1.mha --dimension 512,3
> rtkprojectshepploganphantom -g rot2.xml -o proj2.mha --dimension 512,3
> rtkfdk --dimension 256,1,256 -p . -r proj1.mha -o fdk1.mha -g rot1.xml
> rtkfdk --dimension 256,1,256 -p . -r proj2.mha -o fdk2.mha -g rot2.xml
> clitkImageArithm -i fdk1.mha -j fdk2.mha -o fdk.mha
> ```
> 
> where clitkImageArithm is a tool part of [vv](http://vv.creatis.insa-lyon.fr) which simply adds the two volumes. This works perfectly without [1a20fb4](https://github.com/SimonRit/RTK/commit/1a20fb42acd640d4c1f7e00fdac226174eb3884e) so it seems that this patch is not required. There must be something wrong in your input data?

Nothing wrong with the data, I was actually able to process the stacks separately using two different geometry files, I just didn't think about clitkImageArithm to sum the two fdks contributions and focused on having a final weighted projections stack to input to the fdk pipeline.

To be fair I wanted to streamline all the process from projections to final reconstruction, but it is a personal need so I guess this patch is not really required.
The only thing I would keep is the fix for https://github.com/SimonRit/RTK/issues/290

SimonRit:
Thanks for your answer and sorry because it seems I wasn't clear. With the master branch and simulated data, I don't seem to need the fix for #290 (1a20fb4) to have perfect results so I think #290 is not a bug in RTK but that there is a problem with your geometry definition, either the offset is not accurate or the projection origin is not accurate.

GabrieleBelotti:
> Thanks for your answer and sorry because it seems I wasn't clear. With the master branch and simulated data, I don't seem to need the fix for #290 ([1a20fb4](https://github.com/SimonRit/RTK/commit/1a20fb42acd640d4c1f7e00fdac226174eb3884e)) to have perfect results so I think #290 is not a bug in RTK but that there is a problem with your geometry definition, either the offset is not accurate or the projection origin is not accurate.

I tried clitkImageArithm on the phantom data and the incosistency is no longer present.
Still I think that this may have something to do with the spacing of my projections, I will investigate further and update you on my results

SimonRit:
Maybe you should offset your projections by half a pixel by setting this half pixel shift in the origin? Origin is the coordinate of the center of the first pixel. I'm closing the issue and this PR until further updates. Thanks again!

GabrieleBelotti:
> Maybe you should offset your projections by half a pixel by setting this half pixel shift in the origin? Origin is the coordinate of the center of the first pixel. I'm closing the issue and this PR until further updates. Thanks again!

Yes, I confirm the issue was related to the projections not being centered properly! Everything works as intended; there is a shift in intensity between reconstruction using my custom functions and the sum of the images (this is tackled effectively by dividing all pixels intensities in the summed fdk by 2).

PR #293, opened 2020-01-28
BUG: put back missing code for ITKCudaCommon tests

### LINKED ISSUE #287: ITKCudaCommon tests are not built
Hi,

i notices that the ITKCudaCommon test are built only if the CMake variable "ITK_USE_CUDA" is set to true. After some investigation (hopefully with enough depth) I came to the conclusion that this variable is not set. Neither in ITK nor in RTK. In ITK the variable "ITK_USE_GPU" and for RTK the variable "RTK_USE_CUDA" is available.

What was the original intent of "ITK_USE_CUDA"?
Do I miss something or could it be changed to "RTK_USE_CUDA"?

SimonRit:
The code had been deleted by e0d6f1f but the tests had been excluded
until dfc063e. This should now fix #287.


SimonRit:
The patch was not functional, nvcc threw non trivial errors on the dashboard: `nvcc fatal   : Option '--ptx (-ptx)' is not allowed when compiling for multiple GPU architectures`. I gave up and opted for the Armageddon option, see 45af4dd... I'll merge tomorrow if the nightly compilations go fine.

PR #298, opened 2020-02-06
COMP: Future proof enumeration to integer implicit conversion

hjmjohnson:
ITKv5.1 is enforcing more strict enumeration to integer conversion
checking. The changed code will work both before and after the
proposed change by explicitly casting the enumerations to integer
types.


SimonRit:
Thanks a lot!

PR #300, opened 2020-02-12
CHG: Stream the writer instead of the volume

vlibertiaux:
Immediatly writes a chunk of the reconstructed volume to the output rather than gathering all the chunks then writing.  Avoid allocating the memory for the whole volume.

### DIFF
--- a/applications/rtkfdk/rtkfdk.cxx
+++ b/applications/rtkfdk/rtkfdk.cxx
@@ -186,20 +186,12 @@ main(int argc, char * argv[])
   }
 #endif
 
-  // Streaming depending on streaming capability of writer
-  using StreamerType = itk::StreamingImageFilter<CPUOutputImageType, CPUOutputImageType>;
-  StreamerType::Pointer streamerBP = StreamerType::New();
-  streamerBP->SetInput(pfeldkamp);
-  streamerBP->SetNumberOfStreamDivisions(args_info.divisions_arg);
-  itk::ImageRegionSplitterDirection::Pointer splitter = itk::ImageRegionSplitterDirection::New();
-  splitter->SetDirection(2); // Prevent splitting along z axis. As a result, splitting will be performed along y axis
-  streamerBP->SetRegionSplitter(splitter);
-
   // Write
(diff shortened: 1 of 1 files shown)


SimonRit:
Thanks Vincent! It's exactly what I suggested but I have thought a bit more about this. The reason for using the StreamingImageFilter is to be able to chose the dimension for the splitting, I had forgotten about this. This cannot be done for the writer (if I understand correctly, the splitter is currently fixed to the lowest dimension, see [here](https://github.com/InsightSoftwareConsortium/ITK/blob/master/Modules/IO/ImageBase/src/itkImageIOBase.cxx#L938-L959)). The reason for chosing y instead of z is because it will split the projections too, which would be useful to minimize the CUDA memory (that was your initial choice BTW). In general, the volume will fit in CPU memory. I suggest to keep it as it is. If it's important for you, make a simpler change with just one additional option (e.g., `writer_divisions`). Then modify the commit message to ENH instead of CHG (see [here](https://github.com/InsightSoftwareConsortium/ITK/blob/master/CONTRIBUTING.md)). We could also propose to modify ITK to be able to set the splitter in the writer which should not be too complicated.

vlibertiaux:
Hi Simon,
thanks for the explanation, it helps me understand parts of the code :).
It is not a big issue for me.  It was just nice to see that RTK could
reconstruct a volume for which neither the projections nor the volume would
fit in memory, but it shouldn't happen, unless I am not careful about the
memory already in use.

I will nonetheless fix the issue I have with my own "projection splitter",
since it will allow me to distribute the reconstruction load on different
GPUs.  I have started to have a look on how to do that directly in the
code....and it will be a long and difficult process, but one I want to
undertake anyway.

Thank you for your email,

kind regards,
Vincent

Le jeu. 13 févr. 2020 à 22:40, Simon Rit <notifications@github.com> a
écrit :

SimonRit:
Actually, sleeping on it, I think splitting the writer in another direction does not make sense since one wants to write contiguous pieces of memory. One solution would be to write the volume with an image direction which puts y as the last dimension.
Let us know if we can help with your own problem.

Issue #301, opened 2020-02-13
GetArrayFromImage fails


SimonRit:
The following code
```
import numpy as np
import itk
from itk import RTK as rtk

#itk.GetImageFromArray(np.zeros([20,10,30]).astype(np.float32))
ImageType = itk.Image[itk.F,3]
srcdraw = rtk.ConstantImageSource[ImageType].New()
srcdraw.SetSize([20,10,30])
srcdraw.Update()
itk.GetArrayFromImage(srcdraw.GetOutput())
```
produces the error
```
Traceback (most recent call last):
  File "test.py", line 10, in <module>
    itk.GetArrayFromImage(srcdraw.GetOutput())                  
  File "/home/srit/miniconda3/envs/test36/lib/python3.6/site-packages/itkExtras.py", line 257, in GetArrayFromImage
    return _GetArrayFromImage(image_or_filter, "GetArrayFromImage", keep_axes, update)
  File "/home/srit/miniconda3/envs/test36/lib/python3.6/site-packages/itkExtras.py", line 248, in _GetArrayFromImage
    raise RuntimeError("No suitable template parameter can be found.")
RuntimeError: No suitable template parameter can be found.
```
with the latest version produced by Azure pipelines. Uncommenting the commented line makes it work. The official RTK python package on pypi (v2.0.1) does not show this behavior. Any clue @LucasGandel ?

LucasGandel:
Nice catch! I am not sure what changed as it was not required before but this is the way to go.
We need to make sure itk.PyBuffer is loaded before calling GetArrayFromImage, otherwise [itk.PyBuffer.keys()](https://github.com/InsightSoftwareConsortium/ITK/blob/master/Wrapping/Generators/Python/itkExtras.py#L242) contains only the types defined by itkPyBufferRTK, but the ones from itkPyBuffer are missing, resulting in ["No suitable template parameter..."](https://github.com/InsightSoftwareConsortium/ITK/blob/master/Wrapping/Generators/Python/itkExtras.py#L244) error.
In opposition, GetImageFromArray [loads ](https://github.com/InsightSoftwareConsortium/ITK/blob/master/Wrapping/Generators/Python/itkExtras.py#L287)PyBuffer so this is why the subsequent calls to GetArrayFromImage succeed.
[...]

PR #302, opened 2020-02-13
BUG: add ITKBridgeNumPy dependency to load all templates

SimonRit:
Attempts to solve #301


SimonRit:
@LucasGandel This seems to solve #301. I think that this makes sure that ITKBridgeNumPy is loaded before trying to use `GetImageFromArray` or `GetArrayFromImage`. My guess is that this is required for 751335ae0b9ad07dc5e383e085b755618506845b to work but I'd like your opinion on this before merging.

PR #304, opened 2020-02-14
ENH: added Hnc image IO (another Varian file format) with test

SimonRit:
Supersedes and closes #268. Does that seem reasonable to you @GabrieleBelotti?


GabrieleBelotti:
> Supersedes and closes #268. Does that seem reasonable to you @GabrieleBelotti?

Yes it makes total sense!

PR #305, opened 2020-02-17
Create cleaner builds for ITK v 5.1rc2

hjmjohnson:
This performs a series of code modernizations and best practices cleanups.  Several compiler warnings are removed.


SimonRit:
Thanks, I appreciate you took the time to correct these warnings which were not caught by our dashboard. Any suggestion on which setting to use for catching those?

PR #308, opened 2020-02-19
BUG: direction was filled instead of origin

vlibertiaux:
I might be mistaken but I think the wrong variable was used here.


SimonRit:
Oops. Thanks a lot!

PR #312, opened 2020-02-25
CUDA doesn't support C++17, workaround

agravgaard:
Only forward up to the latest supported C++ standard flag, which is currently c++14


SimonRit:
Thanks! In general, I would rather suggest to avoid C++17 if CUDA is not compatible with it. Maybe there are side effects when mixing two standards? For those who want to take the risks, I would suggest to pass the chosen standard for nvcc via the cmake variable `CUDA_NVCC_FLAGS` and slightly modify the code to check if `-std=c++` is already set (instead of `-std=c++${CMAKE_CXX_STANDARD}` now). That would be cleaner in my opinion. I can make a PR in 10 days if you'd like, I won't be available before I'm afraid.

agravgaard:
That could be a more clean solution, I didn't think of that option.
Then one should probably also get an error message from CMake when configuring with `CMAKE_CXX_STANDARD>=17` and `-std=c++14` is not in `CUDA_NVCC_FLAGS` and cuda enabled.

The mixing works well for me. And I can't really think of any differences in C++14 and 17 that could cause errors at runtime. However, it's probably best to be safe by default.

SimonRit:
Sorry, I ended up not working on this one. The new Cuda 11 supports c++17 so I'm closing it.

PR #317, opened 2020-03-16
PERF: reduced size of application tests

acoussat:
[Some tests](https://my.cdash.org/viewTest.php?onlyfailed&buildid=1790664) fail when the code coverage is computed because they result in timeouts. This pull request reduces the data size for these tests, which should hopefully fix the problem.


SimonRit:
Great thanks.

PR #150, opened 2018-02-11
ENH: Add progress tracking to FDK CPU reconstruction

keithoffer:
My attempt at partially fulfilling #114. I don't have any CUDA enabled hardware, so this is only for CPU based reconstruction.

PR #319, opened 2020-03-18
ENH: ProgressReporter for FDK reconstruction

### LINKED ISSUE #114: itk::ProgressReporter are missing from reconstruction filters
[itk::ProgressReporter](https://itk.org/Doxygen/html/classitk_1_1ProgressReporter.html) would allow to observe the progresses, e.g., with a ProgressBar.

acoussat:
This is a reimplementation of #150 for modern RTK (thanks @keithoffer!). The attempt is to fix #114.

My initial attempts was taken from a4f24efb3ec6e39339a70ace70dc828363ff7299. However the result did not work so well. After investigating for a while, it turns out that starting from ITK 5, ["progress is being reported by multi-threaders on behalf of filters which use `DynamicThreadedGenerateData` signature"](https://github.com/InsightSoftwareConsortium/ITK/blob/master/Documentation/ITK5MigrationGuide.md). So I just removed every progress reporting from a4f24efb3ec6e39339a70ace70dc828363ff7299 for filters implementing `DynamicThreadedGenerateData` signature and it seems to work fine. To be honest I haven't exactly understood how ITK manages to do this and what is exactly performed under the hood.

At first I wanted to make iteration reporting available for other filters, but the ones for which it made sense to me to have iteration reporting already do implement `DynamicThreadedGenerateData`, so, as far as I understand, already somehow implement iteration reporting. I am far from familiar with every aspects of RTK so I put aside filters that I don't understand (and that's a lot!). [Still according to the same page](https://github.com/InsightSoftwareConsortium/ITK/blob/master/Documentation/ITK5MigrationGuide.md), we can enable iteration reporting for threaded filters implementing `GenerateData` interface using `itk::ProgressTransformer`, but it remains a little bit blurry to me.

I will now try to make iteration reporting functional for CUDA FDK reconstruction. In the meantime, could you please tell me for which filters you would like to have iteration reporting enabled, and I'll try to implement it. I'll update this pull request accordingly. Thanks!

### DIFF
--- a/include/rtkFDKConeBeamReconstructionFilter.hxx
+++ b/include/rtkFDKConeBeamReconstructionFilter.hxx
@@ -21,6 +21,8 @@
 
 #include "rtkFDKConeBeamReconstructionFilter.h"
 
+#include <itkProgressAccumulator.h>
+
 namespace rtk
 {
 
@@ -110,6 +112,14 @@ FDKConeBeamReconstructionFilter<TInputImage, TOutputImage, TFFTPrecision>::Gener
   typename ExtractFilterType::InputImageRegionType subsetRegion;
   subsetRegion = this->GetInput(1)->GetLargestPossibleRegion();
   unsigned int nProj = subsetRegion.GetSize(Dimension - 1);
+
+  // The progress accumulator tracks the progress of the pipeline
+  // Each filter is equally weighted across all iterations of the stack
+  itk::ProgressAccumulator::Pointer progress = itk::ProgressAccumulator::New();
+  progress->SetMiniPipelineFilter(this);
+  progress->RegisterInternalFilter(m_WeightFilter, (1.0f/3) / itk::Math::ceil(double(nProj) / m_ProjectionSubsetSize));
+  progress->RegisterInternalFilter(m_RampFilter, (1.0f/3) / itk::Math::ceil(double(nProj) / m_ProjectionSubsetSize));
+  progress->RegisterInternalFilter(m_BackProjectionFilter, (1.0f/3) / itk::Math::ceil(double(nProj) / m_ProjectionSubsetSize));
 
   for (unsigned int i = 0; i < nProj; i += m_ProjectionSubsetSize)
   {
--- a/include/rtkProgressCommands.h
+++ b/include/rtkProgressCommands.h
@@ -1,0 +1,105 @@
+/*=========================================================================
+ *
+ *  Copyright RTK Consortium
+ *
+ *  Licensed under the Apache License, Version 2.0 (the "License");
+ *  you may not use this file except in compliance with the License.
+ *  You may obtain a copy of the License at
+ *
+ *         http://www.apache.org/licenses/LICENSE-2.0.txt
+ *
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ *
+ *=========================================================================*/
+
+#ifndef rtkProgressCommands_h
+#define rtkIterationCommands_h
+
+namespace rtk
+{
+
+/** \class ProgressCommand
+ * \brief Abstract superclass to all progress callbacks.
+ * Derived classes must implement the Run() method.
+ *
+ * \author Aurélien Coussat
+ *
+ * \ingroup RTK
+ *
+ */
+template <typename TCaller>
+class ProgressCommand : public itk::Command
+{
+protected:
+  /** Callback function to be redefined by derived classes. */
+  virtual void
+  Run(const TCaller * caller) = 0;
+
+public:
+  /** Standard class typedefs. */
+  typedef ProgressCommand         Self;
+  typedef itk::Command            Superclass;
+  typedef itk::SmartPointer<Self> Pointer;
+
+  void
+  Execute(itk::Object * caller, const itk::EventObject & event) override
+  {
+    Execute((const itk::Object *)caller, event);
+  }
+
+  void
+  Execute(const itk::Object * caller, const itk::EventObject & event) override
+  {
+    if (!itk::ProgressEvent().CheckEvent(&event))
+    {
+      return;
+    }
+    const auto * cCaller = dynamic_cast<const TCaller *>(caller);
+    if (cCaller)
+    {
+      Run(cCaller);
+    } // TODO fail when cast fails
+  }
+};
+
+/** \class PercentageProgressCommand
+ * \brief Outputs every time a filter progresses by at least one percent.
+ *
+ * \author Aurélien Coussat
+ *
+ * \ingroup RTK
+ *
+ */
+template <typename TCaller>
+class PercentageProgressCommand : public ProgressCommand<TCaller>
+{
+public:
+  /** Standard class typedefs. */
+  typedef PercentageProgressCommand Self;
+  typedef ProgressCommand<TCaller>  Superclass;
+  typedef itk::SmartPointer<Self>   Pointer;
+  itkNewMacro(Self);
+
+  int percentage = -1;
+
+protected:
+  void
+  Run(const TCaller * caller) override
+  {
+    int newPercentage = (int)(caller->GetProgress() * 100.);
+    if (newPercentage > percentage)
+    {
+      std::cout << caller->GetNameOfClass() << " " << newPercentage << "% completed."
+                << std::endl; // TODO allow string personnalization
+      percentage = newPercentage;
+    }
+  }
+};
+
+} // end namespace rtk
+
+#endif
--- a/src/rtkCudaFDKBackProjectionImageFilter.cxx
+++ b/src/rtkCudaFDKBackProjectionImageFilter.cxx
@@ -24,6 +24,7 @@
 #include <itkImageRegionIteratorWithIndex.h>
 #include <itkLinearInterpolateImageFunction.h>
 #include <itkMacro.h>
+#include <itkProgressReporter.h>
 
 namespace rtk
 {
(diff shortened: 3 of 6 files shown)


SimonRit:
Thanks for the work, it's very good news that we don't have to do anything for filters which use `DynamicThreadedGenerateData`. If you want to read more on what this is, see [here](https://itk.org/ITKSoftwareGuide/html/Book1/ITKSoftwareGuide-Book1ch8.html#x54-1380008.4).
About the commit: we could argue about the equal 1/3 weighting for each of the 3 filters in the pipeline but that's not important. I don't understand `nProj / m_ProjectionSubsetSize + 1` in the denominator. The number of subsets is something like `itk::Math::ceil(double(nProj)/m_ProjectionSubsetSize)` I believe.
All filters that don't have `DynamicThreadedGenerateData` need to implement the progress. I think that a PR which works for CPU and Cuda FDK, that's already good a good start. All Cuda filters need to be implemented so for Cuda FDK, you need to implement the progress in the three filters of the mini-pipeline, see [here](https://github.com/SimonRit/RTK/blob/master/include/rtkCudaFDKConeBeamReconstructionFilter.h#L65-L67).

acoussat:
Thanks for the detailed answer. I am giving a look at implementing progress reporting for CUDA FDK now, but I think I need some advice. I dug down the code of `rtkCudaFDKConeBeamReconstructionFilter`, and as far as I understand the GPU kernel [`kernel_weight_projection`](https://github.com/SimonRit/RTK/blob/master/src/rtkCudaFDKWeightProjectionFilter.cu#L33) is the core function actually performing the pixel-wise operations. It looks like that at this point in code we already left all ITK context and we are running pure C-style code, so I don't imagine `itk::ProgressReporter` performing fine here. It's not even in a regular for-loop. For these reasons I don't see a good way to write progress reporting. I tried to explore a little bit the CUDA documentation and I thought I'd be able to plug a thread-wise callback to `kernel_weight_projection`, but I didn't find any way to do this.
Any idea?..

SimonRit:
Indeed, there is no way to report the progress in a cuda kernel. When there is a loop calling the function defined in the `.hcu` file, e.g. in the [backprojection](https://github.com/SimonRit/RTK/blob/master/include/rtkCudaBackProjectionImageFilter.hxx#L101), then you can report other progresses than 0% and 100%. Otherwise, as in [rtkCudaFDKWeightProjectionFilter.cxx](https://github.com/SimonRit/RTK/blob/master/src/rtkCudaFDKWeightProjectionFilter.cxx#L114-L123) it's hopeless, just report 0 and 100% (if it's not already the default behavior inherited from the mother class).

acoussat:
I did my best at implementing this, but the result is kind of disappointing.

- `rtk::CudaFDKWeightProjectionFilter` only reports 0% and 100%, as you mentioned.
- `rtk::CudaFFTRampImageFilter` has the same problem ([no loop around the CUDA call](https://github.com/SimonRit/RTK/blob/master/include/rtkCudaFFTProjectionsConvolutionImageFilter.hxx#L133)).
- `rtk::CudaFDKBackProjectionImageFilter` seems to [only have a slab-wise precision around the CUDA call](https://github.com/SimonRit/RTK/blob/master/src/rtkCudaFDKBackProjectionImageFilter.cxx#L106-L116). By the way, the lines you pointed me to don't seem to refer to the correct loop, since the CUDA call lies below, or am I mistaking? Also, why do we need both `rtk::CudaFDKBackProjectionImageFilter` and `rtk::CudaBackProjectionImageFilter`? You pointed me to the second one but I assumed you wanted to refer to the first one.

With my small tests I'm getting an output for 0, 33, 66 and 100%, which shows that filters do report 0 and 100% by default, which is a good new. Otherwise, I'm not sure how to refine this. If anyone has any pointer, I'd gladly listen!

src/rtkCudaFDKConeBeamReconstructionFilter.cxx
```diff
@@ -38,5 +38,13 @@ rtk::CudaFDKConeBeamReconstructionFilter ::CudaFDKConeBeamReconstructionFilter()
...
 rtk::CudaFDKConeBeamReconstructionFilter ::GPUGenerateData()
 {
+  // The progress accumulator tracks the progress of the pipeline
+  // Each filter is equally weighted across all iterations of the stack
+  itk::ProgressAccumulator::Pointer progress = itk::ProgressAccumulator::New();
+  progress->SetMiniPipelineFilter(this);
+  progress->RegisterInternalFilter(m_WeightFilter, 1.0f/3);
+  progress->RegisterInternalFilter(m_RampFilter, 1.0/3);
+  progress->RegisterInternalFilter(m_BackProjectionFilter, 1.0f/3);
+
```
SimonRit: Is this required? Because it seems to me that this is already (better) implemented in the CPUSuperclass::GenerateData() which is called in this function.

acoussat: Makes sense, I removed those lines. However `rtkCudaFDKConeBeamReconstructionFilter`  was only reporting from 0 to 33%, so I have also added "single-pixel reporting" using `itk::ProgressReporter progress(this, 0, 1)` (the filter has to report only once in order to finish) to `rtkCudaFFTProjectionsConvolutionImageFilter` and `itkCudaFDKWeightProjectionFilter`. I was therefore expecting `rtkCudaFDKConeBeamReconstructionFilter`to jump from 0 to 66%, then report correctly from 66 to 100%, because of the way the mini-pipeline is defined. However, for some reason, the filter correctly reports every single progress from 0 to 100% ?? I don't know exactly what is going on, but it seems to work (which is generally _not_ a good thing).

SimonRit: Great. Remember you still call the GenerateData of the superclass. Therefore, you still divide the projection stacks in pieces and loop over them, see [here](https://github.com/SimonRit/RTK/blob/master/include/rtkFDKConeBeamReconstructionFilter.hxx#L114-L133). The jumps only occur for a subset, not for the full stack.

src/rtkCudaFDKBackProjectionImageFilter.cxx
```diff
@@ -103,6 +103,9 @@ CudaFDKBackProjectionImageFilter ::GPUGenerateData()
       fMatrix[j + (iProj - iFirstProj) * 12] = matrix[j / 4][j % 4];
   }
 
+  // Slab-wise progress reporting
+  itk::ProgressReporter progress(this, 0, itk::Math::ceil(double(nProj) / SLAB_SIZE));
```
SimonRit, line 108: It seems you're missing an include, see [here](https://dev.azure.com/simonrit/RTK/_build/results?buildId=218&view=logs&j=64a1ac24-9467-5512-5818-77318bce79b4&t=7ce6600f-5450-5bf7-9459-cde6129c68fd).

include/rtkProgressCommands.h
```diff
@@ -0,0 +1,104 @@
...
+
+  int percentage = -1;
+
+protected:
+  void
+  Run(const TCaller * caller) override
+  {
+    int newPercentage = (int) (caller->GetProgress() * 100.);
+    if(newPercentage > percentage)
+      {
```
SimonRit: Check style, I don't think it's correct here.

SimonRit (commented):
I guess you can create one commit to finalize the PR after the suggested corrections. Thanks a lot!

src/rtkCudaFDKConeBeamReconstructionFilter.cxx
acoussat: True, I should have guessed. Thanks for the explanation!

src/rtkCudaFDKBackProjectionImageFilter.cxx
acoussat: Added `#include <itkProgressReporter>`, thanks!

include/rtkProgressCommands.h
acoussat: Indeed, forgot to apply `clang-format`, fixed!

include/rtkFDKConeBeamReconstructionFilter.hxx
```diff
@@ -111,6 +113,14 @@ FDKConeBeamReconstructionFilter<TInputImage, TOutputImage, TFFTPrecision>::Gener
   subsetRegion = this->GetInput(1)->GetLargestPossibleRegion();
   unsigned int nProj = subsetRegion.GetSize(Dimension - 1);
 
+  // The progress accumulator tracks the progress of the pipeline
+  // Each filter is equally weighted across all iterations of the stack
+  itk::ProgressAccumulator::Pointer progress = itk::ProgressAccumulator::New();
+  progress->SetMiniPipelineFilter(this);
+  progress->RegisterInternalFilter(m_WeightFilter, (1.0f/3) / itk::Math::ceil(double(nProj) / m_ProjectionSubsetSize));
```
SimonRit, line 120: I'm having second thoughts on ` itk::Math::ceil(double(nProj) / m_ProjectionSubsetSize)`. Maybe replace with `nProj/m_ProjectionSubsetSize+(nProj%m_ProjectionSubsetSize?1:0)`. Or maybe you see sg more adequate to know the times the for loop is going to be executed?

acoussat: My initial guess would have been `nProj/m_ProjectionSubsetSize+(nProj%m_ProjectionSubsetSize?1:0)`, but I don't see how it differs from `itk::Math::ceil(double(nProj) / m_ProjectionSubsetSize)`. Can you imagine any case where they result to different values?
Otherwise I personally find `itk::Math::ceil(double(nProj) / m_ProjectionSubsetSize)` more elegant.

SimonRit: I'm concerned by the precision of double division, can we trust that double(16)/4 will be 4. and not 4.000001... Any information on that?

acoussat: That's a tricky question... I tried to look it up but couldn't find a definitive answer. I played a bit with a toy program,
```c
#include <stdio.h>

int main()
{
	int n = 0x7FFFFFFE; // highest even int
	int d = 2;
	printf("%f", ((double)n) / d);
	return 0;
}
```
but couldn't get any result ending with `...0001` or something similar.

My conclusion is: it probably doesn't matter for such an unimportant part of the program, but I understand that you want to get things right, so I can change it to the other formula you suggested. Do you agree?

SimonRit: Ok, I think it won't occur. I tried a more systematic test but couldn't find a result neither:
```c
#include <stdio.h>

int main()
{
  int d = 80; 
  for(int n=0; n>=0; n+=d) // Stops when it overflows
  {
    if(((double)n) / d -n/d!=0.)
    {
      printf("%d\n", n);                                                                                                                                                                                           
      printf("%a\n", ((double)n) / d -n/d);
      return 1;
    }
  }
  return 0;
}
```
I'll merge as is.

Issue #270, opened 2019-11-22
Python crashes when geometry is not set in some filters


SimonRit:
All filters using the geometry object should test that it's adequately set and issue an exception if not.

acoussat:
Is there a list of filters that should be fixed somewhere? I would like to fix that but I don't feel like testing every single filter...

SimonRit:
I would use something like `grep Geometry include/*.h | grep itkSet | sed "s/:.*$//g"`. I recently discovered that this sort of things should be implemented in `VerifyPreconditions` in ITK filters. Check for examples in the ITK code...

SimonRit:
As far as I know, this issue is solved thanks to @acoussat. I'm closing it

PR #324, opened 2020-04-03
ENH: filters check whether their geometry is set

acoussat:
Attempt at solving #270.

As suggested, this affects all filters returned by `grep Geometry include/*.h | grep itkSet | sed "s/:.*$//g"` as of 72878637c67c55a709c3841bfd51cf0cdce8f035. I suspect that some already existing checks are to be moved in `VerifyPreconditions` but I guess that's something we can leave for future pull requests.

With this version, `rtkAmsterdamShroudTest` fails because of [those lines](https://github.com/acoussat/RTK/blob/check-geometry/test/rtkamsterdamshroudtest.cxx#L230-L231). I don't really know what this filter does and if it is supposed to work without geometry. Can someone explain if this test should be left as it currently is? If yes, I can remove the geometry check.

### DIFF
--- a/include/rtkADMMTotalVariationConeBeamReconstructionFilter.h
+++ b/include/rtkADMMTotalVariationConeBeamReconstructionFilter.h
@@ -214,6 +214,10 @@ protected:
   ADMMTotalVariationConeBeamReconstructionFilter();
   ~ADMMTotalVariationConeBeamReconstructionFilter() override = default;
 
+  /** Checks that inputs are correctly set. */
+  void
+  VerifyPreconditions() ITKv5_CONST override;
+
   /** Does the real work. */
   void
   GenerateData() override;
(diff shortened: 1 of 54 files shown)


SimonRit:
Looks great. Please remove the superfluous lines found with `grep -r "GetGeometry() == nullptr" -A 3`. And for the Amsterdam shroud, the geometry is not mandatory, see the test (and the comment) [here](https://github.com/SimonRit/RTK/blob/master/include/rtkAmsterdamShroudImageFilter.hxx#L113-L115). So undo the modifications of rtkAmsterdamShroudImageFilter.h*. I'll merge when all this has been changed! Thanks a lot, Simon

acoussat:
Should be alright now, tell me if you agree then I'll squash everything!

SimonRit (approved):
I'll merge after commits are squashed. Thanks!

Issue #253, opened 2019-07-15
No default projectors in terative reconstrution


SimonRit:
With the wrappings, not setting the projectors (back and forward) crashes the python kernel. Default projectors or an error message should be used.

SimonRit:
Fixed by #326

PR #326, opened 2020-04-07
ENH: default projectors for all iterative filters

acoussat:
After a lot of trials and errors, my attempt at solving #253.

All filters inheriting from [`rtk::IterativeConeBeamReconstructionFilter`](http://www.openrtk.org/Doxygen/classrtk_1_1IterativeConeBeamReconstructionFilter.html) now have default forward- and back-projectors and won't produce segmentation faults anymore.

This has not been thoroughly tested, I hope it is alright!

### DIFF
--- a/include/rtkADMMTotalVariationConeBeamReconstructionFilter.h
+++ b/include/rtkADMMTotalVariationConeBeamReconstructionFilter.h
@@ -174,15 +174,6 @@ public:
   using DisplacedDetectorFilterType = rtk::DisplacedDetectorImageFilter<TOutputImage>;
   using GatingWeightsFilterType = rtk::MultiplyByVectorImageFilter<TOutputImage>;
 
-  /** Pass the ForwardProjection filter to the conjugate gradient operator */
-  void
-  SetForwardProjectionFilter(ForwardProjectionType _arg) override;
-
-  /** Pass the backprojection filter to the conjugate gradient operator and to the back projection filter generating the
-   * B of AX=B */
-  void
-  SetBackProjectionFilter(BackProjectionType _arg) override;
-
   /** Pass the geometry to all filters needing it */
   itkSetObjectMacro(Geometry, ThreeDCircularProjectionGeometry);
 
--- a/include/rtkADMMTotalVariationConeBeamReconstructionFilter.hxx
+++ b/include/rtkADMMTotalVariationConeBeamReconstructionFilter.hxx
@@ -97,33 +97,6 @@ ADMMTotalVariationConeBeamReconstructionFilter<TOutputImage,
 
 template <typename TOutputImage, typename TGradientOutputImage>
 void
-ADMMTotalVariationConeBeamReconstructionFilter<TOutputImage, TGradientOutputImage>::SetForwardProjectionFilter(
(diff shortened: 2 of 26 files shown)


SimonRit:
Thanks. I only had a quick look and I'm sure it works but my feeling is that only the `SetForwardProjectionFilter` and `SetBackProjectionFilter` of `rtk::IterativeConeBeamReconstructionFilter` should be kept. The code in the daughter classes should be moved to the `GenerateOutputInformation`. This is a bit of refactoring but this is more adequate I believe... What do you think?

acoussat:
Something like this? (I have only updated `rtk::ConjugateGradientConeBeamReconstructionFilter` and if it suits you, I will propagate the modifications to the other classes).

SimonRit:
Yes exactly. Now I wonder if it's not better to let the unknown default values for the fp and the bp, and raise an exception if it's not been set. Because as it is now, one can still set it back to unknown and get a crash I believe.

acoussat:
We could do both: keep default values (useful for quick tests or if the user doesn't know what to pick) and raise an exception if the filter was set back to unknown.
A more drastic solution would be to keep default values and completely remove the values `FP_UNKNOWN` and `BP_UNKNOWN`. What do you think? Do they have any purpose besides indicating that the projectors must absolutely be set?

SimonRit:
I think that only the drastic solution makes sense if we want to keep default values. Try implementing this then.

acoussat:
Tell me if this is fine to you, then I'll squash everything!

SimonRit:
It looks good overall (quick glance), thanks for the work. Trying to see if there is a difference between the two implementations, I see one: the forward/back projectors are always recreated, even if we haven't changed the types. It's a bit hard to evaluate if it's going to make a difference... That's avoidable but makes the implementation a bit less readable. Let's assume it's ok, verify clang-format, squash and I'll merge.

acoussat:
From what I understand, the mother class, `rtk::IterativeConeBeamReconstructionFilter`, [never instantiates its own filters](https://github.com/acoussat/RTK/blob/master/include/rtkIterativeConeBeamReconstructionFilter.hxx#L86-L108) but merely keeps track of what filters are wanted (through `m_CurrentForwardProjectionConfiguration` and `m_CurrentForwardProjectionConfiguration`). I'd say the filters are thus only created once, when `GenerateOutputInformation` is called. Or am I wrong?

SimonRit:
That's correct. Now take a conjugate gradient alternating with denoising. Then the same conjugate gradient algorithm is called multiple times with different inputs so GenerateOutputInformation is recalled every time. Anyway, let's assume that it makes no difference, just squash and I'll merge. Thanks again!

acoussat:
I see, I hadn't thought about that. That's something we would fix later. Do you have any suggestion on how to fix that? Should we create an issue?
On another topic, the branch has been squashed.

SimonRit:
Without a lot of thinking, one potential solution: pass the pointer by reference to `InstantiateBackProjectionFilter` and `InstantiateForwardProjectionFilter` instead of returning it and only create it if it's not already of the right type. Let's leave it as is for the time being and let's wait if we find it's indeed an issue.

PR #328, opened 2020-04-08
ENH: Add rtkSimulatedGeometry python application

LucasGandel:


applications/rtksimulatedgeometry/rtkSimulatedGeometry.py
```diff
@@ -0,0 +1,51 @@
+#!/usr/bin/env python
+import argparse
+from itk import RTK as rtk
+
+if __name__ == '__main__':
+  # Argument parsing
+  parser = argparse.ArgumentParser(description=
```
LucasGandel: @SimonRit I chose to use argparse to avoid having to pip install external packages for argument parsing. I find it clear enough but let me know if you prefer another one (click?).

setup.py
```diff
@@ -40,6 +40,9 @@
         "Operating System :: Unix",
         "Operating System :: MacOS"
         ],
+    scripts=[
+        "lib/rtkSimulatedGeometry.py"
```
LucasGandel: This makes scripts available in the path and allows users to call rtkSimulatedGeometry.py from anywhere just after pip installing rtk.

applications/rtksimulatedgeometry/CMakeLists.txt
```diff
@@ -10,5 +10,9 @@ if(NOT RTK_INSTALL_NO_EXECUTABLES)
       LIBRARY DESTINATION ${RTK_INSTALL_LIB_DIR} COMPONENT RuntimeLibraries
       ARCHIVE DESTINATION ${RTK_INSTALL_ARCHIVE_DIR} COMPONENT Development)
   endforeach()
+
+  # Install Python application
+  install(FILES rtkSimulatedGeometry.py
```
LucasGandel: This installs the application script next to other wrapped files

LucasGandel:
@SimonRit This is how we can provide python applications to users. Let me know what you think.

applications/rtksimulatedgeometry/rtkSimulatedGeometry.py
SimonRit: Ok sounds like a reasonable choice.

applications/rtksimulatedgeometry/CMakeLists.txt
SimonRit: In general, I still have the feeling that the cmake `install` commands in RTK are not right and should be replaced by, e.g., `itk_module_target_install`. But for this case I did not find a good macro in [ITKModuleMacros.cmake](https://github.com/InsightSoftwareConsortium/ITK/blob/master/CMake/ITKModuleMacros.cmake).

setup.py
SimonRit: Great!

applications/rtksimulatedgeometry/rtkSimulatedGeometry.py
```diff
@@ -0,0 +1,51 @@
+#!/usr/bin/env python
```
SimonRit: Do you mind renaming it to rtksimulatedgeometry.py for consistency? Is `#!/usr/bin/env python` cross-platform?

SimonRit (changes requested):
Unfortunately, I can't test the generated package because setup.py refers to ITK v5.1. Do you mind changing it to v5.1rc03 please? Thanks!

applications/rtksimulatedgeometry/rtkSimulatedGeometry.py
LucasGandel: Renaming is done. I tried it on windows and it works fine, might be just ignored

applications/rtksimulatedgeometry/CMakeLists.txt
LucasGandel: I understand, maybe such macro could help to improve RTK install process. But for this specific use case I think it is completely fine.

applications/rtksimulatedgeometry/rtksimulatedgeometry.py
```diff
@@ -0,0 +1,51 @@
...
+import argparse
+from itk import RTK as rtk
+
+if __name__ == '__main__':
+  # Argument parsing
+  parser = argparse.ArgumentParser(description=
+    "Creates an RTK geometry file from simulated/regular trajectory. See http://www.openrtk.org/Doxygen/DocGeo3D.html for more information.")
+
+  parser.add_argument('nproj', metavar='n', type=int, help='Number of projections')
+  parser.add_argument('output', metavar='output', help='Output file name')
```
SimonRit: Why are those not similar to the c++ application (--nproj and --output)?
My current feeling is that we should mirror the applications.

applications/rtksimulatedgeometry/rtksimulatedgeometry.py
```diff
@@ -0,0 +1,51 @@
+#!/usr/bin/env python
+import argparse
```
SimonRit, line 2: If I anticipate the translation of all ggo files, how easy do you think it will be to have an "include" mechanism for ggo files such as [rtk3Doutputimage_section.ggo](https://github.com/SimonRit/RTK/blob/master/applications/rtk3Doutputimage_section.ggo)?

SimonRit (changes requested):
Looks good and it seems to work although there are a bunch of warnings when loading the RTK package. These warnings are unrelated to the PR and I'll handle them, please don't do it, I'm in the middle of a large modification of the wrappings.

applications/rtksimulatedgeometry/rtksimulatedgeometry.py
LucasGandel: Very good question! I tried to find a mechanism to avoid duplicating the options but I could not find anything easy so far. The problem is to bring the included files in the package.
What I can think of is a mechanism similar to the one we have with CMake in C++, that would automatically parse and generate .py files to be included by the application script later to define the args object.
I can try to setup this and add it to this PR, unless you think of something else?

applications/rtksimulatedgeometry/rtksimulatedgeometry.py
LucasGandel: Agreed! But missing "--" indicates that it is a required argument for the script to run.
If those argument are not specified, the application prints the help section. It avoids having to check for the validity of those arguments in the code, but if you prefer to have only optional arguments I can change that.

applications/rtksimulatedgeometry/rtksimulatedgeometry.py
SimonRit: I guess the alternative is a small python module containing a function with the parser object as a parameter. Nothing urgent, we can leave it for later.

applications/rtksimulatedgeometry/rtksimulatedgeometry.py
SimonRit: As far as I understand, according to `rtksimulatedgeometry.py -h`, they become positional arguments:
```
positional arguments:
  n                     Number of projections
  output                Output file name
```
And indeed, `rtksimulatedgeometry.py -n 4 -o test.xml` gives `rtksimulatedgeometry.py: error: unrecognized arguments: -n -o`. So, I don't know how to use these positional arguments and yes, I prefer to pass them as optional arguments. Similarly, I would provide a default value for --rad_cyl and always pass the value. The result is the same, but it makes it clearer that no option is equivalent to rad_cyl=0.

applications/rtksimulatedgeometry/rtksimulatedgeometry.py
LucasGandel: I tried a few things as discussed on Friday, but could not find a clean solution so far.
I think we can leave it like this for now and improve it later depending on what is decided for C++ application (keep them or not).
Other changes about arguments have been addressed.

SimonRit:
Thanks again! To be continued...

PR #330, opened 2020-04-10
Clang format linter

LucasGandel:
@SimonRit This should run clang-format on PR and make the pipeline fail if the style is inconsistent. For branches pushed on the repo (not fork), style can automatically be fixed by adding the label action:ApplyClangFormat

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,15 +1,20 @@
-.git*            export-ignore
+.git*                                   export-ignore
 # Custom attribute to mark sources as using our C++/C code style.
 [attr]our-c-style  whitespace=tab-in-indent,no-lf-at-eof  hooks.style=KWStyle,clangformat,uncrustify
 
-
-*.c              our-c-style
-*.h              our-c-style
-*.cxx            our-c-style
-*.hxx            our-c-style
-*.txx            our-c-style
-*.txt            whitespace=tab-in-indent,no-lf-at-eof
-*.cmake          whitespace=tab-in-indent,no-lf-at-eof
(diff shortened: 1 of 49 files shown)


.gitattributes
```diff
@@ -1,15 +1,19 @@
...
-*.hxx            our-c-style
-*.txx            our-c-style
-*.txt            whitespace=tab-in-indent,no-lf-at-eof
-*.cmake          whitespace=tab-in-indent,no-lf-at-eof
+include/*.h*                            our-c-style
+src/*.c*                                our-c-style
+applications/**/*.h                     our-c-style
+applications/**/*.cxx                   our-c-style
+applications/**/*.hxx                   our-c-style
+examples/*.cxx                          our-c-style
```
SimonRit: Shouldn't this be `examples/**/*.cxx`?

SimonRit (changes requested):
Ok thanks. Can you rebase your 3 commits on the updated #329? I have included the cmake/*cxx files, maybe you want to do the same. There is also one comment to address.
Locally, will this automatically apply the style as is done on a commit in the ITK repository?

.gitattributes
LucasGandel: Nice catch! Thanks!

LucasGandel:
@SimonRit I made the changes you requested, thanks.
The linter will run automatically on each PR as soon as this is merged. 
For automatically fixing the style, you must add the "action:ApplyClangFormat" label to your PR. But this works only for branches pushed to your repository (not forks).
I am still working on the hooks to fix style locally but this is an additional check.

PR #331, opened 2020-04-10
BUG: solve memory leak in rtkcheckimagequality

acoussat:
`rtkcheckimagequality` does not use the `GGO` macro hence causes memory leaks. This pull request solves that.


SimonRit:
Thanks!

Issue #151, opened 2018-02-21
Implement an option in SART to set the threshold of DivideOrZeroOutImageFilter


cyrilmory:
See this thread for details : https://public.kitware.com/pipermail/rtk-users/2018-February/010575.html
Chao Wu tested this solution and it worked for his data

SimonRit:
Fixed by #333

PR #333, opened 2020-04-14
ENH: set division threshold for SART

acoussat:
My attempt at fixing #151. From what I understood of the problem, this should suffice, but I might be missing something. I guess the modifications are automatically propagated to the Python wrappings.
One remark though: the current implementation allows setting a negative threshold. Is that a problem? [ITK seems to allow this behavior.](https://github.com/InsightSoftwareConsortium/ITK/blob/master/Modules/Filtering/ImageIntensity/include/itkDivideOrZeroOutImageFilter.h#L74-L84)

### DIFF
--- a/applications/rtksart/rtksart.cxx
+++ b/applications/rtksart/rtksart.cxx
@@ -119,6 +119,11 @@ main(int argc, char * argv[])
     sart->SetEnforcePositivity(true);
   }
 
+  if (args_info.divisionthreshold_given)
+  {
+    sart->SetDivisionThreshold(args_info.divisionthreshold_arg);
+  }
+
   REPORT_ITERATIONS(sart, rtk::SARTConeBeamReconstructionFilter<OutputImageType>, OutputImageType)
 
   TRY_AND_EXIT_ON_ITK_EXCEPTION(sart->Update())
--- a/include/rtkSARTConeBeamReconstructionFilter.h
+++ b/include/rtkSARTConeBeamReconstructionFilter.h
@@ -147,6 +147,7 @@ public:
   /** Some convenient type alias. */
   using VolumeType = TVolumeImage;
   using ProjectionType = TProjectionImage;
+  using ProjectionPixelType = typename ProjectionType::PixelType;
 
   /** Typedefs of each subfilter of this composite filter */
   using ExtractFilterType = itk::ExtractImageFilter<ProjectionType, ProjectionType>;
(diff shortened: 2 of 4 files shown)


applications/rtksart/rtksart.cxx
```diff
@@ -119,6 +119,11 @@ main(int argc, char * argv[])
     sart->SetEnforcePositivity(true);
   }
 
+  if (args_info.divisionthreshold_given)
+    {
```
SimonRit: Check clang-format...

include/rtkSARTConeBeamReconstructionFilter.h
```diff
@@ -209,6 +210,13 @@ class ITK_EXPORT SARTConeBeamReconstructionFilter
   itkSetMacro(DisableDisplacedDetectorFilter, bool);
   itkGetMacro(DisableDisplacedDetectorFilter, bool);
 
+  /** Set the threshold below which pixels in the denominator in the projection space are considered zero. The division
+   * by zero will then be evaluated at zero. Avoid noise magnification from low projections values when working with
+   * noisy and/or simulated data.
+   */
+  void
+  SetDivisionThreshold(ProjectionPixelType threshold);
```
SimonRit: Please use ITK macros `itkSetMacro itkGetMacro` for set and get (which is missing).

include/rtkSARTConeBeamReconstructionFilter.hxx
```diff
@@ -119,6 +119,13 @@ SARTConeBeamReconstructionFilter<TVolumeImage, TProjectionImage>::SetGatingWeigh
   m_IsGated = true;
 }
 
+template <class TVolumeImage, class TProjectionImage>
+void
+SARTConeBeamReconstructionFilter<TVolumeImage, TProjectionImage>::SetDivisionThreshold(ProjectionPixelType threshold)
+{
+  m_DivideProjectionFilter->SetThreshold(threshold);
```
SimonRit: I understand now why you haven't used ITK macros. I used to do similar things and RTK is full of such strategies. I now think it's not good because if you change the threshold value, the filter is not modified and the filter won't update if you call `Update()` again. My suggestion: create a `m_DivisionThreshold` filter parameter, Set Get it with the ITK macros and pass it to the filter in the `GenerateOutputInformation`. That's the most readable solution I believe.

SimonRit (changes requested):
Well done, I had no idea it was that simple. Please address my comments and we're done.

applications/rtksart/rtksart.cxx
acoussat: Woops, I don't understand why `clang-format` didn't apply correctly on this file. Done!

acoussat:
Updated my commit. Hope everything is correct now!

PR #336, opened 2020-04-16
ENH: Add quadratic penalization to OSEM algorithm

arobert01:
Implement the quadratic penalization describes in [De Pierro, IEEE TMI,
1995].

### DIFF
--- a/include/rtkOSEMConeBeamReconstructionFilter.hxx
+++ b/include/rtkOSEMConeBeamReconstructionFilter.hxx
@@ -40,6 +40,7 @@ OSEMConeBeamReconstructionFilter<TVolumeImage, TProjectionImage>::OSEMConeBeamRe
   m_NumberOfIterations = 3;
   m_SigmaZero = -1;
   m_Alpha = -1;
+  m_BetaRegularization = 0.01;
 
   // Create each filter of the composite filter
   m_ExtractFilter = ExtractFilterType::New();
@@ -47,6 +48,7 @@ OSEMConeBeamReconstructionFilter<TVolumeImage, TProjectionImage>::OSEMConeBeamRe
   m_ConstantVolumeSource = ConstantVolumeSourceType::New();
   m_ZeroConstantProjectionStackSource = ConstantProjectionSourceType::New();
   m_DivideProjectionFilter = DivideProjectionFilterType::New();
+  m_DePierroRegularizationFilter = DePierroRegularizationFilterType::New();
 
   // Create the filters required for the normalization of the
   // backprojection
@@ -55,7 +57,7 @@ OSEMConeBeamReconstructionFilter<TVolumeImage, TProjectionImage>::OSEMConeBeamRe
 
   // Permanent internal connections
   m_DivideProjectionFilter->SetInput1(m_ExtractFilter->GetOutput());
-  m_MultiplyFilter->SetInput1(m_DivideVolumeFilter->GetOutput());
+  m_DivideVolumeFilter->SetInput1(m_MultiplyFilter->GetOutput());
 
   // Default parameters
   m_ExtractFilter->SetDirectionCollapseToSubmatrix();
@@ -185,11 +187,15 @@ OSEMConeBeamReconstructionFilter<TVolumeImage, TProjectionImage>::GenerateOutput
 
   m_BackProjectionNormalizationFilter->SetTranspose(false);
 
-  m_DivideVolumeFilter->SetInput1(m_BackProjectionFilter->GetOutput());
-  m_DivideVolumeFilter->SetInput2(m_BackProjectionNormalizationFilter->GetOutput());
-  m_DivideVolumeFilter->SetConstant(0);
-
+  m_MultiplyFilter->SetInput1(m_BackProjectionFilter->GetOutput());
   m_MultiplyFilter->SetInput2(this->GetInput(0));
+
+  m_DePierroRegularizationFilter->SetInput(0, this->GetInput(0));
+  m_DePierroRegularizationFilter->SetInput(1, m_MultiplyFilter->GetOutput());
+  m_DePierroRegularizationFilter->SetInput(2, m_BackProjectionNormalizationFilter->GetOutput());
+  m_DePierroRegularizationFilter->SetBeta(m_BetaRegularization);
+  m_DivideVolumeFilter->SetInput2(m_DePierroRegularizationFilter->GetOutput());
+  m_DivideVolumeFilter->SetConstant(0);
 
   m_ForwardProjectionFilter->SetInput(0, m_ZeroConstantProjectionStackSource->GetOutput());
   m_ForwardProjectionFilter->SetInput(1, this->GetInput(0));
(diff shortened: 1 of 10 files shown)


include/rtkDePierroRegularizationFilter.h
```diff
@@ -0,0 +1,159 @@
...
+
+#ifndef rtkDePierroRegularizationFilter_h
+#define rtkDePierroRegularizationFilter_h
+
+#include <itkImageToImageFilter.h>
+#include <itkSquareImageFilter.h>
+#include <itkSqrtImageFilter.h>
+#include <itkMultiplyImageFilter.h>
+#include <itkSubtractImageFilter.h>
+#include <itkAddImageFilter.h>
```
SimonRit: Remove? Seems unused.

include/rtkDePierroRegularizationFilter.h
```diff
@@ -0,0 +1,159 @@
...
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ *
+ *=========================================================================*/
+
+#ifndef rtkDePierroRegularizationFilter_h
+#define rtkDePierroRegularizationFilter_h
+
+#include <itkImageToImageFilter.h>
```
SimonRit: Remove? Seems useless (nearly all filters inherit from it so must include it).

include/rtkDePierroRegularizationFilter.h
```diff
@@ -0,0 +1,159 @@
...
+#include <itkSquareImageFilter.h>
+#include <itkSqrtImageFilter.h>
```
SimonRit: Remove? Seems unused.

include/rtkDePierroRegularizationFilter.h
```diff
@@ -0,0 +1,159 @@
...
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ *
+ *=========================================================================*/
+
+#ifndef rtkDePierroRegularizationFilter_h
+#define rtkDePierroRegularizationFilter_h
+
+#include <itkImageToImageFilter.h>
+#include <itkSquareImageFilter.h>
```
SimonRit: Remove? Seems unused.

include/rtkDePierroRegularizationFilter.h
```diff
@@ -0,0 +1,159 @@
...
+#include <itkSqrtImageFilter.h>
+#include <itkMultiplyImageFilter.h>
+#include <itkSubtractImageFilter.h>
+#include <itkAddImageFilter.h>
+#include <itkDivideImageFilter.h>
```
SimonRit: Remove? Seems unused.

include/rtkDePierroRegularizationFilter.h
```diff
@@ -0,0 +1,159 @@
...
+ * \ingroup RTK ReconstructionAlgorithm
+ */
+template <class TInputImage, class TOutputImage = TInputImage>
+class ITK_EXPORT DePierroRegularizationFilter : public itk::ImageToImageFilter<TInputImage, TOutputImage>
+{
+public:
+  ITK_DISALLOW_COPY_AND_ASSIGN(DePierroRegularizationFilter);
+
+  /** Standard class type alias. */
+  typedef DePierroRegularizationFilter Self;
```
SimonRit: Replace all typedef by using (if we want to follow the new ITK style)...

include/rtkDePierroRegularizationFilter.h
```diff
@@ -0,0 +1,159 @@
...
+  /** Run-time type information (and related methods). */
+  itkTypeMacro(DePierroRegularizationFilter, itk::ImageToImageFilter);
+
+  /** Get / Set the hyper parameter for the regularization */
+  itkGetMacro(Beta, float);
+  itkSetMacro(Beta, float);
+
+protected:
+  DePierroRegularizationFilter();
+  ~DePierroRegularizationFilter() ITK_OVERRIDE = default;
```
SimonRit: Prefer `override` over `ITK_OVERRIDE` (actually only your filters still use ITK_OVERRIDE, maybe we should clean this in a separate commit).

include/rtkDePierroRegularizationFilter.hxx
```diff
@@ -0,0 +1,140 @@
...
+#include "rtkDePierroRegularizationFilter.h"
+
+namespace rtk
+{
+template <class TInputImage, class TOutputImage>
+DePierroRegularizationFilter<TInputImage, TOutputImage>::DePierroRegularizationFilter()
+{
+  this->SetNumberOfRequiredInputs(2);
+  // Set default parameters
+  m_Beta = 0.01;
```
SimonRit: Prefer initialization where it's declared (see e.g., 819c8a6).

include/rtkDePierroRegularizationFilter.hxx
```diff
@@ -0,0 +1,140 @@
...
+  m_DefaultNormalizationVolume = ConstantVolumeSourceType::New();
+  m_SubtractImageFilter = SubtractImageFilterType::New();
+  m_ConvolutionFilter = NOIFType::New();
+  m_CustomBinaryFilter = CustomBinaryFilterType::New();
+
+  // Set Lambda function
+  auto customLambda = [](const typename InputImageType::PixelType & input1,
+                         const typename InputImageType::PixelType & input2) -> typename OutputImageType::PixelType
+  {
+    return static_cast<typename OutputImageType::PixelType>((input1 + std::sqrt(input1 * input1 + input2)) / 2);
```
SimonRit: Nice!

SimonRit:
Looks great, I have added a few style suggestions to follow the elvolving ITK style. Can you add a simple test? I guess you only need to set a beta in rtkosemtest.cxx...

SimonRit (changes requested):

arobert01:
By default the regularization is applied with a hyperparameter of 0.01, so rtkosemtest.cxx take already into account the regularization. Would you prefer to set the hyperparameter to 0 by default ?

include/rtkDePierroRegularizationFilter.h
arobert01: You are right I will do it in a next style commit.

include/rtkDePierroRegularizationFilter.hxx
arobert01: Maybe I could do it also in a separate commit and apply it to my others filters as well, what do you think ?

include/rtkDePierroRegularizationFilter.h
arobert01: Addressed by 9245de6

SimonRit:
> By default the regularization is applied with a hyperparameter of 0.01, so rtkosemtest.cxx take already into account the regularization. Would you prefer to set the hyperparameter to 0 by default ?

Sorry, I missed that. It's fine like this, thanks!

include/rtkDePierroRegularizationFilter.hxx
arobert01: Done with 99497e1

include/rtkDePierroRegularizationFilter.h
```diff
@@ -0,0 +1,155 @@
...
+ *  CustomBinary -> Output;
+ *  }
+ * \enddot
+ *
+ * \author Antoine Robert
+ *
+ * \ingroup RTK ReconstructionAlgorithm
+ */
+template <class TInputImage, class TOutputImage = TInputImage>
+class ITK_EXPORT DePierroRegularizationFilter : public itk::ImageToImageFilter<TInputImage, TOutputImage>
```
SimonRit: Sorry, I missed that in my previous review but maybe the name should be DePierroRegularizationImageFilter? All the rest looks good, thanks a lot!

arobert01: Yes sorry my bad ... The last force-pushed fixed this !

include/rtkOSEMConeBeamReconstructionFilter.hxx
```diff
@@ -223,23 +222,23 @@ OSEMConeBeamReconstructionFilter<TVolumeImage, TProjectionImage>::GenerateOutput
   }
 
   m_DivideProjectionFilter->SetInput2(m_ForwardProjectionFilter->GetOutput());
-  m_DivideProjectionFilter->SetConstant(0);
+  m_DivideProjectionFilter->SetConstant(1);
```
SimonRit, line 225: Out of curiosity, why do you set the constant here and line 191 since it seems to me that you just want to do the division between two images?

SimonRit (commented):
I'll merge as is if everything passes except if you want to fix the two typos (describe -> described and change -> changed) + if you feel you can remove those SetConstant which seem superfluous.

include/rtkOSEMConeBeamReconstructionFilter.hxx
arobert01: Here, the filter used is [itk::DivideOrZeroOutImageFilter](https://github.com/InsightSoftwareConsortium/ITK/blob/master/Modules/Filtering/ImageIntensity/include/itkDivideOrZeroOutImageFilter.h), the constant in this case is the value returned when the denominator input value is considered zero.

arobert01:
I will fix the typo, sorry.

PR #338, opened 2020-04-19
BUG: Change member type from float to double in Zeng projectors

arobert01:
This should fix the bug introduced by 9ec2bc5d53307e47ffd60f8fa5d8102d3315a5f8.


SimonRit:
Not there yet: https://my.cdash.org/viewBuildError.php?buildid=1810483

arobert01:
Should be good now

PR #339, opened 2020-04-20
ENH: Compute the backprojection of ones only once in OSEM

arobert01:
A backprojection of a volume filled with ones is needed for the OSEM algorithm.
Until now, this backprojection was computed at every iteration. However,
the result is independent of the iterations. Thus, computing this
backprojection only during the first iteration and reused the result for
the next is enough and should reduce the computation time of the OSEM
algorithm.

### DIFF
--- a/include/rtkOSEMConeBeamReconstructionFilter.hxx
+++ b/include/rtkOSEMConeBeamReconstructionFilter.hxx
@@ -257,6 +257,11 @@ OSEMConeBeamReconstructionFilter<TVolumeImage, TProjectionImage>::GenerateData()
   // Should be tunable with other solutions.
   std::vector<unsigned int> projOrder(nProj);
 
+  // If m_StoreNormalizationImages is true, the backprojection of
+  // ones will be only computed once during the first iteration.
+  // The result will be stored in an vector and reused for the next iterations.
+  std::vector<typename TVolumeImage::Pointer> vectorNorm;
+
   for (unsigned int i = 0; i < nProj; i++)
     projOrder[i] = i;
   std::shuffle(projOrder.begin(), projOrder.end(), Superclass::m_DefaultRandomEngine);
@@ -271,6 +276,7 @@ OSEMConeBeamReconstructionFilter<TVolumeImage, TProjectionImage>::GenerateData()
   for (unsigned int iter = 0; iter < m_NumberOfIterations; iter++)
   {
     unsigned int projectionsProcessedInSubset = 0;
+    unsigned int currentSubset = 0;
     for (unsigned int i = 0; i < nProj; i++)
     {
       // Change projection subset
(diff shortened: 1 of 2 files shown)


include/rtkOSEMConeBeamReconstructionFilter.hxx
```diff
@@ -271,37 +278,45 @@ OSEMConeBeamReconstructionFilter<TVolumeImage, TProjectionImage>::GenerateData()
...
+      }
 
       projectionsProcessedInSubset++;
       if ((projectionsProcessedInSubset == m_NumberOfProjectionsPerSubset) || (i == nProj - 1))
       {
+        if (iter == 0)
+        {
+          duplicator->SetInputImage(m_BackProjectionNormalizationFilter->GetOutput());
+          duplicator->Update();
+          vectorNorm.push_back(duplicator->GetOutput());
```
SimonRit: I think I'm missing something but what I would have tried is
```
          vectorNorm.push_back(m_BackProjectionNormalizationFilter->GetOutput());
          vectorNorm.back()->DisconnectPipeline();
```
to avoid the need for using the duplicator. What am I missing?

include/rtkOSEMConeBeamReconstructionFilter.hxx
```diff
@@ -21,6 +21,7 @@
 
 #include "rtkOSEMConeBeamReconstructionFilter.h"
 #include "rtkGeneralPurposeFunctions.h"
+#include "itkImageDuplicator.h"
```
SimonRit: If you do another change, consider using `<>` instead of `""` for itk includes.

include/rtkOSEMConeBeamReconstructionFilter.hxx
```diff
@@ -261,6 +262,12 @@ OSEMConeBeamReconstructionFilter<TVolumeImage, TProjectionImage>::GenerateData()
     projOrder[i] = i;
   std::shuffle(projOrder.begin(), projOrder.end(), Superclass::m_DefaultRandomEngine);
 
+  // The backprojection of ones will be only computed once during the first iteration.
+  // The result will be stored in an vector and reused for the next iterations.
+  std::vector<typename TVolumeImage::Pointer> vectorNorm;
```
SimonRit: Probably quite memory consuming. Should we consider making this optional?

SimonRit (commented):
Great, thanks. Can you look at my comments?

include/rtkOSEMConeBeamReconstructionFilter.hxx
SimonRit: One more thought: maybe you should store the inverse and do a multiplication instead of a division. That should be more computationally efficient.

include/rtkOSEMConeBeamReconstructionFilter.hxx
arobert01: No, you are right ! I didn't think about this.

include/rtkOSEMConeBeamReconstructionFilter.hxx
arobert01: > Probably quite memory consuming. Should we consider making this optional?

I'am not sure to understand what you mean, would you prefer to keep the way it was before (which is more time consuming) ? Or do you suggest to find a way to do it without using a std::vector ? 

  

> One more thought: maybe you should store the inverse and do a multiplication instead of a division. That should be more computationally efficient.

The result of this back-projection is used by  [rtk::DePierroRegularizationImageFilter](https://github.com/SimonRit/RTK/blob/adb8061fe951f46477078d828959ca6533392d79/include/rtkDePierroRegularizationImageFilter.h) to compute the regularization factor. Thus, if I store the inverse I would also need to inverse the formula of the regularization factor computation.

SimonRit: I suggest to add an option `bool m_StoreNormalizationImages` and to let the user chose between memory (new version) and computational time (old version).
Ok for the other comment, I was not aware of this.

arobert01: Ok good idea, I'll do it and I will amend my previous commit.

include/rtkOSEMConeBeamReconstructionFilter.hxx
```diff
@@ -327,9 +343,12 @@ OSEMConeBeamReconstructionFilter<TVolumeImage, TProjectionImage>::GenerateData()
...
-        m_BackProjectionNormalizationFilter->SetInput(0, norm);
+        if (iter == 0 || !m_StoreNormalizationImages)
+        {
+          norm = m_BackProjectionNormalizationFilter->GetOutput();
+          norm->DisconnectPipeline();
+          m_BackProjectionNormalizationFilter->SetInput(0, norm);
+        }
       }
     }
     this->GraftOutput(pimg);
```
SimonRit, line 354: Looks good. Should you call `vectorNorm.clear()` at the end?

arobert01: Done

PR #342, opened 2020-04-24
Pre commit clang

LucasGandel:


LucasGandel:
@SimonRit This forces users to run utilities/SetupForDevelopment.sh to be allowed to commit.
KWStyle and ClangFormat are downloaded at build time and run on pre-commit.
Clangformat fixes are applied but must be reviewed with git-mergetool so it requires users to setup their git-mergetool. A warning about how to do so is thrown when trying to commit.

SimonRit:
Sounds perfect!

PR #343, opened 2020-04-24
COMP: Make sure GIT_DIR is set in pre-commit

LucasGandel:


LucasGandel:
@SimonRit This is updating the hooks branch for #342 to work

PR #344, opened 2020-04-24
Include PyBuffer.i at a submodule level

LucasGandel:
Include additional swig library files at a submodule level to avoid
mixing swig %import and %include directives.
See the original discourse discussion [1] for a description of the
issue.
It is possible to append additional swig includes in the submodule's
_ext.i file that is only included by the corresponding submodule.
This must be done between the itk_wrap_submodule_python() and
itk_end_wrap_submodule_python() where the _ext.i is created, so using
the submodule .wrap file makes sense.
Using this approach, RTK can mirror the logic from ITK PyBuffer* files.
The previous approach using WRAPPER_SWIG_LIBRARY_FILES can be kept to
define additional includes at a module level.

.[1]:https://discourse.itk.org/t/additional-python-types-for-itk-remote-external-modules/2509


LucasGandel:
@SimonRit @arobert01 This improves PyBuffer wrapping by including the PyBuffer.i only in the submodule. It allows us to mirror ITK's PyBuffer.init/PyBuffer.i approach hence making maintenance easier.

SimonRit:
Thanks! If I check PyBufferRTK.i.in, I think the difference with PyBuffer.i.in on the ITK repository can be proposed there and, if accepted and released, PyBufferRTK.i.in can be removed from the RTK repository. For PyBuffer.i.init, I think I would rather remove it, do the configure_file with the one from the ITK source directory and use a cmake commande to add `from itkImageDuplicatorRTKPython import *`. These two suggestions avoid the need for merging changes on the ITK side to RTK. What do you think?

LucasGandel:
Agreed, that would be ideal.
The problem is to get access to ITK's PyBuffer.i* files when building python packages against ITKPythonBuilds. We don't have access to ITK sources there. But I guess we could fetch the two files directly from github using CMake. I'll give it a try and let you know.

SimonRit:
Can we compile RTK without access to ITK sources? I'm a bit surprised...

LucasGandel:
Ahah right... they must be somewhere, I'll check what's best to access the two files.

LucasGandel:
@SimonRit  I updated the change to use ITK PyBuffer files. Please do not merge until antoine's change are integrated into ITK.

@arobert01 Do you have time to add you change regarding PyBuffer's dimension 1 into ITK?
You only need to add the two lines, [l. 107](https://github.com/SimonRit/RTK/commit/72ed4dfb7fdf294c4ca56e47088a7e76dc00807d#diff-27d4c4e4912c44e121d03d9d1472465aR107) and [l. 118](https://github.com/SimonRit/RTK/commit/72ed4dfb7fdf294c4ca56e47088a7e76dc00807d#diff-27d4c4e4912c44e121d03d9d1472465aR118). (Also maybe update the comment you forgot [here](https://github.com/SimonRit/RTK/commit/72ed4dfb7fdf294c4ca56e47088a7e76dc00807d#diff-27d4c4e4912c44e121d03d9d1472465aL108)).

arobert01:
> @arobert01 Do you have time to add you change regarding PyBuffer's dimension 1 into ITK?
> You only need to add the two lines, [l. 107](https://github.com/SimonRit/RTK/commit/72ed4dfb7fdf294c4ca56e47088a7e76dc00807d#diff-27d4c4e4912c44e121d03d9d1472465aR107) and [l. 118](https://github.com/SimonRit/RTK/commit/72ed4dfb7fdf294c4ca56e47088a7e76dc00807d#diff-27d4c4e4912c44e121d03d9d1472465aR118). (Also maybe update the comment you forgot [here](https://github.com/SimonRit/RTK/commit/72ed4dfb7fdf294c4ca56e47088a7e76dc00807d#diff-27d4c4e4912c44e121d03d9d1472465aL108)).

Yes I will do it, I'll let you know when it's done.

SimonRit:
@LucasGandel, when @arobert01 PR is merged, do we need to wait for a new realease of ITKPythonPackage before merging this PR?

arobert01:
It is merged !

LucasGandel:
@SimonRit I just checked [ITKPythonBuilds](https://github.com/InsightSoftwareConsortium/ITKPythonBuilds/releases) and you are right, they were released before @arobert01 changes. So we will have to wait for a new release of the builds otherwise we'll loose support for dimension 1.

SimonRit (approved):
I'll merge as soon as ITK releases a new version.

SimonRit:
InsightSoftwareConsortium/ITK#1806 will be in the next release (InsightSoftwareConsortium/ITK@caae21695f6d7402bc1596404c59ca983df44183) so let's merge, we can still compile without it.

PR #346, opened 2020-05-04
Setup Sphinx documentation

LucasGandel:
@SimonRit Here is a POC regarding automated online documentation.

This is far from being complete but it sets up the minimum infrastructure to automatically generate local and online documentation.
We could keep pushing code to this PR until we have a complete documentation gathering all the resources already available online.

I tried to replicate the wiki applications pages into the examples directory. This cannot be built nor tested with CMake but it makes doc updates easier.

Support for existing markdown files has been added. An example is given by including GettingStarted.md in the main index.rst file.

Support for external images has been added. They can be added to any directory. The documentation/Sphinx/ExternalData folder is a good place to store images that should be included in the doc but that don't belong to any example.

An example of the resulting doc should be available [here](https://lucasgandelrtk.readthedocs.io/en/latest/).

We can setup a call to review and discuss this PR together, just let me know.


LucasGandel:
@SimonRit Images should now be fixed in the documentation.
I could keep the original approach:
- Images can be added using "content-files" referencing the sha512 from data.kitware. (see examples/FDK)
- Images are automatically downloaded when building the doc.
- Images must be referenced using relative path when included in .rst files (see [here](https://github.com/SimonRit/RTK/pull/346/files#diff-01f10fc9c2334707f2382a5f28643a5cR17-R19))

We can schedule a call to discuss this again.

SimonRit:
Thanks, it looks good. We now need to have an internal discussion to agree (or not) on this documentation format... and to write it.

LucasGandel:
PR updated, a preview of the generated doc is available here: https://lucasgandelrtk.readthedocs.io/en/latest/

A README file has been added to share instructions on how to extend the doc.

Note that the top level index.md file must be placed next to the conf.py file. Because we want the documentation files to be spread across the sources, both index.md and conf.py must be at the root of the RTK source tree.

GettingStarted.md
```diff
@@ -5,8 +5,8 @@ Here are suggested steps for the RTK beginner.
 
   1. Install the software following [INSTALLATION.md](INSTALLATION.md).
 
-  2. If you are not already familiar with ITK, [get started with ITK](https://github.com/InsightSoftwareConsortium/ITK/blob/master/GettingStarted.md).
+  2. If you are not already familiar with ITK, [get started with ITK](https://github.com/InsightSoftwareConsortium/ITK/blob/master/GettingStarted.md#L1).
```
SimonRit: Why this `#L1`?

conf.py
```diff
@@ -0,0 +1,66 @@
...
+# -- Project information -----------------------------------------------------
+project = 'RTK'
+copyright = f'{date.today().year}, Simon Rit'
+author = 'RTK Consortium'
+
+# The full version, including alpha/beta/rc tags
+# release = '2.6.0'
+
+# -- General configuration ---------------------------------------------------
+extensions = [
```
SimonRit, line 25: The `.readthedocs.yml` file is the same as ITK's but I see quite a few differences in the `extensions` (`'sphinx.ext.napoleon', 'sphinx.ext.intersphinx', 'sphinxext.opengraph', 'sphinx_design'` are not on the RTK side). Shouldn't we make the two files as close as possible for maintanability?

documentation/docs/README.md
```diff
@@ -0,0 +1,57 @@
...
+
+Thank you for improving our documentation!
+
+The sources of the RTK documentation are spread across the RTK source tree in multiple markdown (.md) files.
+Edit existing documentation files, add new examples or documentation files, and open a merge request on the RTK repository.
+
+## Add a new documentation page
+To add a new page to the documentation, create a new .md file in the RTK source tree.
+Write plain markdown text or use the [MyST](https://myst-parser.readthedocs.io/en/latest/index.html) syntax to include technical and scientific content.
+The MyST extension enabled when building the documentation are defined by the `myst_enable_extensions` variable in the sphinx `conf.py` configuration file.
```
SimonRit: I would remove this line (or simplify it to "The MyST extension is enabled when building the documentation" but this is implied by the previous sentence).

SimonRit (approved):
This is great, I only have minor suggestions. I will split examples into "Code examples" and "Applications examples" with a top page explaining the difference but that's up to me to do it. Thanks @LucasGandel!
Any preference between merging as is or taking over the PR to add more content?

GettingStarted.md
LucasGandel: This is useless here, it corresponds to the line number. I can remove it

conf.py
LucasGandel: Agreed, I will synchronize the extensions on both sides. These extensions are not used in the RTK doc (yet?), but I don't think it hurts to add them

documentation/docs/README.md
LucasGandel: I agree the line is misleading. The idea here is that MyST defines its own extensions, for instance `attrs_inline` to add inline images. Using the MyST syntax is not enough and the corresponding extension must be added to the `myst_enable_extensions` variable in the sphinx `conf.py` configuration file.

Should I remove the line or replace it with something like: "Additional MyST syntax entensions can be enabled by adding them to the `myst_enable_extensions` variable in the sphinx `conf.py` configuration file" ?

SimonRit: Sounds good to me!

conf.py
SimonRit: If you think it's better to keep them minimal keep it as is.

documentation/docs/README.md
SimonRit: Sounds good to me!

conf.py
LucasGandel: OK I kept it as is, we can always synchronize this later if we see that other extensions are needed

LucasGandel:
> This is great, I only have minor suggestions. I will split examples into "Code examples" and "Applications examples" with a top page explaining the difference but that's up to me to do it. Thanks @LucasGandel! Any preference between merging as is or taking over the PR to add more content?

Suggestions have been addressed. I think it is better to merge this now and setup the readthedocs build with your account. Once it is up and running, we can make changes to the doc in new MRs

PR #355, opened 2020-06-12
BUG: fix iteration reporting for 4D conjugate gradient

SimonRit:
It would be nice to check whether #266 similarly missed other iterative algorithms @acoussat


acoussat:
I have ran some tests on my side. Your fix seems to work, thanks for noticing! I didn't notice any additional problem. However, as an easy way to test that every iterative filters' reporting works as expected is lacking, I cannot guarantee that everything is alright. Please tell me if any other filter does not behave as wanted, I'll fix it as fast as I can!

PR #358, opened 2020-06-23
ENH: one-line iteration reporting

acoussat:
Should enhance iteration reporting outputs such that it only occupies a single line.

However this approach has some drawbacks:
- if for some reason the filter outputs something, then this output is going to be erased after iteration completion;
- the character `\r` is output when there is a redirection to a file, which produces a line like
```
^MIteration 1 completed.^MIteration 2 completed.^MIteration 3 completed.
```
- a new line is missing at the very end of the output.

I don't see any potential fix for these issues, tell me if an idea pops to your mind!

### DIFF
--- a/applications/rtkfdk/rtkfdk.cxx
+++ b/applications/rtkfdk/rtkfdk.cxx
@@ -150,7 +150,7 @@ main(int argc, char * argv[])
   f->SetProjectionSubsetSize(args_info.subsetsize_arg);                                                                \
   if (args_info.verbose_flag)                                                                                          \
   {                                                                                                                    \
-    f->AddObserver(itk::ProgressEvent(), progressCommand);                                                             \
+    f->AddObserver(itk::AnyEvent(), progressCommand);                                                                  \
   }
 
   // FDK reconstruction filtering
--- a/include/rtkMacro.h
+++ b/include/rtkMacro.h
@@ -215,7 +215,7 @@
   {                                                                                                                    \
     using VerboseIterationCommandType = rtk::VerboseIterationCommand<filter_type>;                                     \
     typename VerboseIterationCommandType::Pointer verboseIterationCommand = VerboseIterationCommandType::New();        \
-    filter->AddObserver(itk::IterationEvent(), verboseIterationCommand);                                               \
+    filter->AddObserver(itk::AnyEvent(), verboseIterationCommand);                                                     \
   }                                                                                                                    \
   if (args_info.output_every_given)                                                                                    \
   {                                                                                                                    \
@@ -233,6 +233,5 @@
     filter->AddObserver(itk::IterationEvent(), outputIterationCommand);                                                \
   }
 //--------------------------------------------------------------------
-
 
 #endif
--- a/include/rtkIterationCommands.h
+++ b/include/rtkIterationCommands.h
@@ -37,17 +37,6 @@ namespace rtk
 template <typename TCaller>
 class IterationCommand : public itk::Command
 {
-protected:
-  /** How many times this command has been executed. */
-  unsigned int m_IterationCount = 0;
-
-  /** Trigger the callback every n iterations. */
-  unsigned int m_TriggerEvery = 1;
-
-  /** Callback function to be redefined by derived classes. */
-  virtual void
-  Run(const TCaller * caller) = 0;
-
 public:
   /** Standard class typedefs. */
   typedef IterationCommand        Self;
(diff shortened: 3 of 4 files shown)


SimonRit:
Thanks. Of the 3 points, only the last one is a problem in my opinion. Maybe we should use a `ProgressReporter` in addition to the `IterationReporter` and catch the end of this progress reporter to output a end of line. Does that seem possible? Maybe it's simpler to create an IterationAndProgressReporter to report both at the same time.

acoussat:
I see some drawbacks using this approach. Filters only subscribe to one type of event at a time, which means that filters should subscribe to both `itk::ProgressReporter` and `itk::IterationReporter` to correctly output the line as supposed. This is actually not really a problem because that complexity can be hidden behind the macro [`REPORT_ITERATIONS`](https://github.com/SimonRit/RTK/blob/master/include/rtkMacro.h#L213). But that means that every filter should also implement `ProgressReporter`. It is feasible for filters whose iteration count is known before execution, but is it the case for all iterative filters? Otherwise they could report a 0-100% progress, but it might seem a bit overkill for such a minor feature.
(On a side not, it would perhaps be interesting that filters having a fixed iteration count also report their progress in percentage using `ProgressReporter`, what do you think?)

Another solution would be to use charcodes to move up one line after printing a new line, for instance [using `\e[A`](https://stackoverflow.com/a/3277094), but it definitely won't be cross-platform.

SimonRit:
> it might seem a bit overkill for such a minor feature.

I agree, maybe it's not worth all your efforts. Consider closing this PR if you feel it's too much work.

>  (On a side not, it would perhaps be interesting that filters having a fixed iteration count also report their progress in percentage using ProgressReporter, what do you think?)

That's what I had in mind, yes.

> Another solution would be to use charcodes to move up one line after printing a new line, for instance using \e[A, but it definitely won't be cross-platform.

If it's not cross-platform, it's not an option.

applications/rtkadmmtotalvariation/rtkadmmtotalvariation.cxx
```diff
@@ -19,6 +19,7 @@
 #include "rtkadmmtotalvariation_ggo.h"
 #include "rtkGgoFunctions.h"
 #include "rtkConfiguration.h"
+#include "rtkEndCommands.h"
```
SimonRit: Great! I would just move all these includes to a single one in rtkMacro.h.

acoussat:
I am noticing that the classes `IterationCommand`, `ProgressCommand` and `EndCommand` are almost similar, I think it would be great to merge them into a single class. Now that I think about it, it would even be possible to write a command that acts as follow:
- if the `itkEventObject` is of type `ProgressEvent` or `IterationEvent`, output the update;
- if the `itkEventObject` is of type `EndEvent`, output a new line.

Does that seem better to you?

include/rtkMacro.h
```diff
@@ -231,8 +231,10 @@
       outputIterationCommand->SetFileFormat("iter%d.mha");                                                             \
     }                                                                                                                  \
     filter->AddObserver(itk::IterationEvent(), outputIterationCommand);                                                \
-  }
+  }                                                                                                                    \
+  using NewLineEndCommandType = rtk::NewLineEndCommand<filter_type>;                                                   \
```
SimonRit: Shouldn't there be a  `if (args_info.verbose_flag)`?

SimonRit:
> Does that seem better to you?

Yes!

acoussat:
Alright, latest commit should fix everything. In my opinion the code is indeed cleaner now!

applications/rtkfdk/rtkfdk.cxx
```diff
@@ -150,7 +150,7 @@ main(int argc, char * argv[])
   f->SetProjectionSubsetSize(args_info.subsetsize_arg);                                                                \
   if (args_info.verbose_flag)                                                                                          \
   {                                                                                                                    \
-    f->AddObserver(itk::ProgressEvent(), progressCommand);                                                             \
+    f->AddObserver(itk::AnyEvent(), progressCommand);                                                                  \
```
SimonRit, line 153: Some useless spaces on this line after ;

include/rtkMacro.h
```diff
@@ -215,7 +215,7 @@
   {                                                                                                                    \
     using VerboseIterationCommandType = rtk::VerboseIterationCommand<filter_type>;                                     \
     typename VerboseIterationCommandType::Pointer verboseIterationCommand = VerboseIterationCommandType::New();        \
-    filter->AddObserver(itk::IterationEvent(), verboseIterationCommand);                                               \
+    filter->AddObserver(itk::AnyEvent(), verboseIterationCommand);                                                     \
```
SimonRit, line 218: Idem

SimonRit (approved):
Yes I agree, it looks good. You can just remove some extra spaces in two places. Can you squash and I'll merge? Thanks again for all the work.

acoussat:
I don't see how these are useless spaces? They keep the backslashes aligned to the right, as I thought was expected from the code style. Or perhaps I am missing something?

SimonRit:
> I don't see how these are useless spaces? They keep the backslashes aligned to the right, as I thought was expected from the code style. Or perhaps I am missing something?

My bad, I misread the diff. Just squash then, Sorry and thanks!

acoussat:
Force-pushed squashed commits, thanks!

PR #362, opened 2020-07-02
ENH: progress reporting for ramp filter

acoussat:
While running some tests using `rtkfdk -v`, I noticed that the reported progress ended at 66%. After some investigation, it turned out that the ramp filter didn't report its progress anymore, probably because of a719d8913 which removed the use of `DynamicThreadedGenerateData`. This pull request brings back progress reporting for ramp filter, along with reported progress using `rtkramp -v`.

I don't know if the code is correct, since the exact working of `itk::ProgressAccumulator` is somewhat still blurry to me, so please do tell me if anything is wrong!


SimonRit:
Thanks and sorry I broke it.

PR #363, opened 2020-07-03
COMP: Support if(... IN_LIST...) with CMake old projects

finetjul:
If a project with cmake_minimum_required(VERSION 3.0) (e.g. CTK)
find_packages ITK that includes RTK built with CUDA, then
if(... IN_LIST ...) in FindCUDA_wrap.cmake is not supported.

We do not want to enforce Foo to have the same miminum version as RTK.
Foo could add policy but find_package(ITK) creates a new policy scope that prevents Foo from
ignores policies added by Foo.


SimonRit:
Thanks. Before merging, I'd like to address #312 at the same time, I hope to do that before the end of the week

PR #364, opened 2020-07-03
COMP: Prevent RTK from overwriting CMAKE_MODULE_PATH

finetjul:
The existing behavior was erasing foo from CMAKE_MODULE_PATH while find_package(ITK) was executed:
```
set(CMAKE_MODULE_PATH foo)
find_package(ITK)
```

This is because _ITKConfig.cmake_ eventually called _lib/cmake/Modules/RTK.cmake_ that had `CMAKE_MODULE_PATH` already expanded when ITK was built.


SimonRit:
Thanks!

PR #366, opened 2020-07-07
Fixed detector offset initialization

sbrousmiche:
Uninitialized variable fix for centered detector


SimonRit:
We have now switched to ITK format for the commit messages (see [here)](https://github.com/InsightSoftwareConsortium/ITK/blob/master/CONTRIBUTING.md). Do you mind doing a `git commit --amend` and add a `BUG:` prefix then `git push -f` ?

PR #370, opened 2020-07-23
ENH: Configure out pixel value in FieldOfViewImageFilter

finetjul:


SimonRit:
Looks good, thanks. For consistency with [BinaryThresholdImageFilter](https://itk.org/Doxygen/html/classitk_1_1BinaryThresholdImageFilter.html) (and the obvious pleasure of typing long variable names), should it be Inside and Outside (instead of In and Out)? No filter uses `InValue` and `OutValue` in ITK...

finetjul:
sounds good, I'm updating it.

SimonRit:
Perfect, thanks.

Issue #296, opened 2020-01-30
ITKCudaCommon is missing tests


SimonRit:
Testing is indirectly done in RTK but that would be good to have specific tests. Some old tests have been removed by 45af4dd since fc70a70 raised the error
```
nvcc fatal   : Option '--ptx (-ptx)' is not allowed when compiling for multiple GPU architectures
```
See discussion in #287.

SimonRit:
There's been one added with RTKConsortium/ITKCudaCommon@ec97d5bd60ad9309785b7f8bc7b35e376a70d8c4, further work needed to be reported in [RTKConsortium/ITKCudaCommon](https://github.com/RTKConsortium/ITKCudaCommon).

PR #371, opened 2020-07-29
WIP: Add FFTHilbertImageFilter

acoussat:
First attempt at creating a filter to compute the Hilbert transform of an image.

A few questions:
- there are no tests of this new filter; should I write some?
- the kernel is a band-limited Hilbert kernel, but there are no ways in the current implementation to choose that band-limit, should that be an option?

Thanks!

### DIFF
--- a/include/rtkFFTHilbertImageFilter.h
+++ b/include/rtkFFTHilbertImageFilter.h
@@ -1,0 +1,112 @@
+/*=========================================================================
+ *
+ *  Copyright RTK Consortium
+ *
+ *  Licensed under the Apache License, Version 2.0 (the "License");
+ *  you may not use this file except in compliance with the License.
+ *  You may obtain a copy of the License at
+ *
+ *         http://www.apache.org/licenses/LICENSE-2.0.txt
+ *
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ *
+ *=========================================================================*/
+
+#ifndef rtkFFTHilbertImageFilter_h
+#define rtkFFTHilbertImageFilter_h
+
+#include <itkConceptChecking.h>
+#include "rtkConfiguration.h"
+#include "rtkFFTProjectionsConvolutionImageFilter.h"
+#include "rtkMacro.h"
+
+namespace rtk
+{
+
+/** \class FFTHilbertImageFilter
+ * \brief Implements the Hilbert transform.
+ *
+ * \test rtkhilbertfiltertest.cxx
+ *
+ * \author Aurélien Coussat
+ *
+ * \ingroup RTK ImageToImageFilter
+ */
+
+template <class TInputImage, class TOutputImage = TInputImage, class TFFTPrecision = double>
+class ITK_EXPORT FFTHilbertImageFilter
+  : public rtk::FFTProjectionsConvolutionImageFilter<TInputImage, TOutputImage, TFFTPrecision>
+{
+public:
+  ITK_DISALLOW_COPY_AND_ASSIGN(FFTHilbertImageFilter);
+
+  /** Standard class type alias. */
+  using Self = FFTHilbertImageFilter;
+  using Superclass = rtk::FFTProjectionsConvolutionImageFilter<TInputImage, TOutputImage, TFFTPrecision>;
+  using Pointer = itk::SmartPointer<Self>;
+  using ConstPointer = itk::SmartPointer<const Self>;
+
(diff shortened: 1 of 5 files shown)


include/rtkFFTHilbertImageFilter.hxx
```diff
@@ -0,0 +1,89 @@
...
+
+  // Compute band-limited kernel in space domain
+  double    spacing = this->GetInput()->GetSpacing()[0];
+  IndexType ix;
+
+  while (!it.IsAtEnd())
+  {
+    ix = it.GetIndex();
+
+    double x = ix[0] * spacing;
```
SimonRit: Please add a `m_PixelShift` option. This will require redefining the `GenerateOutputInformation()` member.

include/rtkFFTHilbertImageFilter.h
```diff
@@ -0,0 +1,122 @@
...
+
+#ifndef rtkFFTHilbertImageFilter_h
+#define rtkFFTHilbertImageFilter_h
+
+#include <itkConceptChecking.h>
+#include "rtkConfiguration.h"
+#include "rtkFFTProjectionsConvolutionImageFilter.h"
+#include "rtkMacro.h"
+
+// The Set macro is redefined to clear the current FFT kernel when a parameter
```
SimonRit: This is useless in thecurrent code since itkSetMacro is not used. I am suggesting to add a new member, `m_PixelShift`, so it would become useful, but it's actually simpler to directly use the new definition instead of this complicated trick.

SimonRit (changes requested):
In addition to the 2 comments, I think we should remove the application and create a wrap file. I think @acoussat does not have time to finalize this in the coming days, do you need it urgently @mathu59?

acoussat:
I have implemented all your comments, thanks again!

I have also created a test for this new filter, but it might still require some work. Specifically:
- I think I correctly disabled zero padding to write the test, but I don't exactly get the result I am supposed to (the one obtained analytically). Am I supposed to manually mirror the signal out of its current bounds? Does RTK already has an option somewhere to do this?
- I don't really know how to pick good parameters for `CheckImageQuality`, as of now they are nothing but random choices. Any tip about that?

Thanks!

acoussat:
Here is a comparison representing the current state of the test. The signal `t -> sin(20*pi*t)` has the analytic Hilbert transform `t -> -cos(20*pi*t)`. The current implementation kind of matches the analytic inverse, but still seems incorrect.
hilbert_compare
I'm unsure about how to improve that... There is probably something to do with the way zero padding is performed, but I'm not sure I understand. Any pointers?

thewtex:
Hi,

Hopefully a helpful pointer (and a request :-)) for a monogenic signal filter:

https://arxiv.org/abs/1703.09199

which is the multi-dimensional analog to the analytic signal from the Hilbert transform. We have an AnalyticSignalImageFilter here: 

https://github.com/KitwareMedical/ITKUltrasound/blob/master/include/itkAnalyticSignalImageFilter.h

SimonRit:
Thanks. `itkAnalyticSignalImageFilter.h`seems to be doing the same thing as the current `rtkHilbertImageFilter.h` (which was badly named because we followed Matlab). Is this going to be moved to an ITK module at some point?

thewtex:
@SimonRit Yes, we will make ITKUltrasound available as an ITK remote module following @dzenanz 's work to support remote module dependencies in ITK 5.2:

https://github.com/InsightSoftwareConsortium/ITK/commit/fa1082b92953b09c48ae0d65a96ddfbfd33c7c36#diff-e7efda826b6d097e47d7cb2cd6b1cf126084197d7834e7b022048183509c774b

It would be nice to also have ultrasound filters depend on the CUDA filter support in RTK. I am wondering if now is the time to create a CUDA base module, and a module for FFT extensions, e.g. analytic signal, etc. that the ultrasound and RTK modules could share as dependencies. What do you think?

SimonRit:
Sounds good. For CUDA, it might be easy but we should fix #296. For FFT extensions, RTK would certainly benefit from the analytic signal. We also have filters which work line-by-line or slice-by-slice, the mother class is [rtk::FFTProjectionsConvolutionImageFilter](https://github.com/SimonRit/RTK/blob/master/include/rtkFFTProjectionsConvolutionImageFilter.h). They are pretty specific to tomographic reconstruction but they might be useful in other contexts. I'm sure you'll have some changes to suggest on the code to make it better!

acoussat:
I'm not sure I understand the conclusion of this discussion. Do we keep the current implementation of the Hilbert transform? Should I improve it somehow, for instance by including the analytic signal? Should it be moved somewhere, like to a module that contains FFT extensions?

dzenanz:
I understood that I should finish https://github.com/KitwareMedical/ITKUltrasound/pull/123 before proceeding. Then we should take the parts of ITKUltrasound which overlap ones in RTK and put them in a new common module on which both ITKUltrasound and RTK would depend.

PR #365, opened 2020-07-06
COMP: Declare ITKCudaCommon as an IMPORTED library in ITKTargets.cmake

finetjul:
See #360

PR #374, opened 2020-08-25
CMake fixes for ITKCudaCommon

SimonRit:
Supersedes and closes #365


SimonRit:
@finetjul the failing builds of #365 are gone so I believe it can be merged. Can you confirm these changes are still required before I merge?

finetjul:
@SimonRit no, it should not be merged as is. There is an error first time ITK is configured. It is ok second time it gets configured.

This is due to  `itk_module_target_export(ITKCudaCommon)` that expects a target (?) that does not exist yet.

I have not have found the time to fix such error first time ITK is configured though.
(please note that I also have some configuration errors with recent CUDA SDK version, but that can be fixed in a separate MR)

SimonRit:
Closing, as ITKCudaCommon is being moved to a separate repository, https://github.com/SimonRit/ITKCudaCommon.

PR #375, opened 2020-09-08
Fix documentation typo

finetjul:


SimonRit:
It's not really a typo, just another [notation](https://en.wikipedia.org/wiki/Natural_logarithm). Is it really helpful to change this?

finetjul:
Granted, it is not a typo. I just think it is confusing. By using this notation (not used anywhere else in RTK), it can make believe that the other `log` usage is [log10](https://en.wikipedia.org/wiki/Common_logarithm). I therefore prefer avoiding potential confusion.
I'm ok to not merge it though. No big deal.

SimonRit:
Ok, I have amended the commit message and pushed to master directly see a41c6b2. Thanks!

PR #389, opened 2020-11-03
COMP: missing definition of ITK_LIBRARY_BUILD_TYPE for external build

SimonRit:
Without the patch, there were errors when compiling with BUILD_SHARED_LIBS ON and CUDA ON with MSVC 2019, see [here](https://my.cdash.org/viewBuildError.php?buildid=1911065). 
@LucasGandel  can you tell me what you think?
@dzenanz, @thewtex, is it intentional that ITK_LIBRARY_BUILD_TYPE is not defined when building a remote module externally? RTK is not the only module to use this, there are a few others (at least IOFDF, IOMeshSTL, IOTransformDCMTK, MGHIO, MinimalPathExtraction).

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -131,6 +131,12 @@ mark_as_advanced(RTK_DATA_ROOT)
 # Initialization
 if(NOT ITK_SOURCE_DIR)
   include(itk-module-init.cmake)
+
+  if(RTK_BUILD_SHARED_LIBS)
+    set(ITK_LIBRARY_BUILD_TYPE "SHARED")
+  else()
+    set(ITK_LIBRARY_BUILD_TYPE "STATIC")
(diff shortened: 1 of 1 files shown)


dzenanz (approved):

dzenanz:
It is probably not intentional. I am not sure where would be the proper place to define it.

thewtex:
It looks like this was addressed here:

https://github.com/InsightSoftwareConsortium/ITK/commit/54e902400c0b0930aa99ea95a6e57d34956d4fe0

looks good - thanks, @SimonRit !

I am working on the 5.2 release candidate over the next, when we can test this out.

SimonRit:
Thanks for the poke, I'm not sure why I did not merge this but I guess I can now.

PR #401, opened 2021-01-20
ENH: fix export macros usage to match ITK rules

laurentmalka:
- The rules are the following:
    - Use ModuleName_EXPORT for non-templated classes.
    - Use ITK_TEMPLATE_EXPORT for templated classes.
    - Use ITK_FORWARD_EXPORT for forward declarations.

### DIFF
--- a/include/rtkADMMTotalVariationConeBeamReconstructionFilter.h
+++ b/include/rtkADMMTotalVariationConeBeamReconstructionFilter.h
@@ -134,7 +134,7 @@ template <typename TOutputImage,
           typename TGradientOutputImage =
             itk::Image<itk::CovariantVector<typename TOutputImage::ValueType, TOutputImage::ImageDimension>,
                        TOutputImage::ImageDimension>>
-class ADMMTotalVariationConeBeamReconstructionFilter
+class ITK_TEMPLATE_EXPORT ADMMTotalVariationConeBeamReconstructionFilter
   : public rtk::IterativeConeBeamReconstructionFilter<TOutputImage, TOutputImage>
 {
 public:
--- a/include/rtkADMMTotalVariationConjugateGradientOperator.h
+++ b/include/rtkADMMTotalVariationConjugateGradientOperator.h
@@ -104,7 +104,7 @@ template <typename TOutputImage,
           typename TGradientOutputImage =
             itk::Image<itk::CovariantVector<typename TOutputImage::ValueType, TOutputImage::ImageDimension>,
                        TOutputImage::ImageDimension>>
-class ADMMTotalVariationConjugateGradientOperator : public ConjugateGradientOperator<TOutputImage>
+class ITK_TEMPLATE_EXPORT ADMMTotalVariationConjugateGradientOperator : public ConjugateGradientOperator<TOutputImage>
 {
 public:
 #if ITK_VERSION_MAJOR == 5 && ITK_VERSION_MINOR == 1
--- a/include/rtkADMMWaveletsConeBeamReconstructionFilter.h
+++ b/include/rtkADMMWaveletsConeBeamReconstructionFilter.h
@@ -140,7 +140,7 @@ namespace rtk
  */
 
 template <typename TOutputImage>
-class ADMMWaveletsConeBeamReconstructionFilter
+class ITK_TEMPLATE_EXPORT ADMMWaveletsConeBeamReconstructionFilter
   : public rtk::IterativeConeBeamReconstructionFilter<TOutputImage, TOutputImage>
 {
 public:
--- a/include/rtkADMMWaveletsConjugateGradientOperator.h
+++ b/include/rtkADMMWaveletsConjugateGradientOperator.h
@@ -95,7 +95,7 @@ namespace rtk
  */
 
 template <typename TOutputImage>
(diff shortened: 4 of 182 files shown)


laurentmalka:
@LucasGandel

SimonRit:
Thanks, for the clarification of the rules and for the code. Are the rules explained somewhere in the ITK software guide? A link would be useful, it's always been a bit of a mistery to me (also because these exports are useless on Linux).
The change looks good but there are now new warnings:
https://my.cdash.org/viewBuildError.php?type=1&buildid=1954275
I think they should be addressed in a separate commit. Do you mind doing it in the same PR?

SimonRit (approved):

LucasGandel:
Thanks @SimonRit !

Laurent saw the warning and is working on fixing it, I'll review as soon as it is pushed.

The original discussion about export can be found [here](https://discourse.itk.org/t/correct-usage-of-export-macros/1304), maybe we can describe this in some kind of RTK contribution/developper guide. file?

SimonRit:
Thanks, I now see it's already page 364 of the [ItkSoftwareGuide](https://itk.org/ItkSoftwareGuide.pdf), no need to repeat this. The rule is still to follow ITK guidelines in RTK, thus minimizing the RTK developer guide...

SimonRit:
Thanks it works! But I don't like so much the change, we should not have to inline functions for this reason. Have you tried including [itkWin32Header.h](https://github.com/InsightSoftwareConsortium/ITK/blob/master/Modules/Core/Common/include/itkWin32Header.h)? It seems to simply disable the warning...

laurentmalka:
I misunderstood what the warning meant exactly, but you are right, disabling the warning should be perfectly fine as long as the member variables using STL types remain as private or protected fields.  
This should now be fixed.

I noticed the pipeline for my last push failed though, and it seems to have been caused by a timeout. Is there something I need to fix or is that just something that happens sometimes ?

SimonRit:
I think it's still running isn't it?
https://github.com/SimonRit/RTK/actions/runs/503159488
You'll have to change the commit name with a `COMP:` prefix (see [CONTRIBUTING.md](https://github.com/InsightSoftwareConsortium/ITK/blob/master/CONTRIBUTING.md)). Can you add the warnings (or one example) in the commit message for future reference? Thanks again

laurentmalka:
I was talking about this one, which failed  
https://github.com/SimonRit/RTK/actions/runs/501139953

SimonRit:
> I was talking about this one, which failed
> https://github.com/SimonRit/RTK/actions/runs/501139953

Ok thanks. Let's see if it happens to the latest commit. We might have to split the MacOS compilations, but I'd like to propagate such a change to ITKModuleTemplate for consistency.

include/rtkCudaScatterGlareCorrectionImageFilter.h
```diff
@@ -26,6 +26,7 @@
 
 #  include "rtkScatterGlareCorrectionImageFilter.h"
 #  include "rtkCudaFFTProjectionsConvolutionImageFilter.h"
+#  include "RTKExport.h"
```
LucasGandel: Unneeded?

include/rtkDualEnergyNegativeLogLikelihood.h
```diff
@@ -20,6 +20,7 @@
 #define rtkDualEnergyNegativeLogLikelihood_h
 
 #include "rtkProjectionsDecompositionNegativeLogLikelihood.h"
+#include "RTKExport.h"
```
LucasGandel: Unneeded?

include/rtkCudaScatterGlareCorrectionImageFilter.h
LucasGandel: Unneeded?

include/rtkDualEnergyNegativeLogLikelihood.h
LucasGandel: Unneeded?

include/rtkProjectionsDecompositionNegativeLogLikelihood.h
```diff
@@ -23,6 +23,7 @@
 #include <itkVectorImage.h>
 #include <itkVariableLengthVector.h>
 #include <itkVariableSizeMatrix.h>
+#include "RTKExport.h"
```
LucasGandel: Unneeded?

include/rtkSchlomka2008NegativeLogLikelihood.h
```diff
@@ -20,6 +20,7 @@
 #define rtkSchlomka2008NegativeLogLikelihood_h
 
 #include "rtkProjectionsDecompositionNegativeLogLikelihood.h"
+#include "RTKExport.h"
```
LucasGandel: Unneeded?

LucasGandel (approved):
@laurentmalka Thanks! The changes look good!
I added a few "Unneeded?" comments in classes that include rtkExport.h but don't use RTK_EXPORT. Did you forget to call RTK_EXPORT for those classes or is the include unneeded?

SimonRit:
Thanks again!

PR #404, opened 2021-03-01
Implementation of variance reconstruction

jannisdickmann:
Implementation of variance reconstruction with a new FFTVarianceRampImageFilter and modifications to FFTRampImageFilter and FDKWeightProjectionFilter.

### DIFF
--- a/include/rtkFDKWeightProjectionFilter.hxx
+++ b/include/rtkFDKWeightProjectionFilter.hxx
@@ -60,6 +60,11 @@ FDKWeightProjectionFilter<TInputImage, TOutputImage>::BeforeThreadedGenerateData
       m_ConstantProjectionFactor[k] *= itk::Math::abs(sdd) / (2. * sid * sid);
       m_ConstantProjectionFactor[k] *= sp.GetNorm();
     }
+    if(this->m_VarianceMode)
+    {
+      // square if in variance mode
+      m_ConstantProjectionFactor[k] = m_ConstantProjectionFactor[k] * m_ConstantProjectionFactor[k];
+    }
   }
 }
 
--- a/include/rtkFFTRampImageFilter.h
+++ b/include/rtkFFTRampImageFilter.h
@@ -128,6 +128,10 @@ public:
   itkGetConstMacro(SheppLoganCutFrequency, double);
   itkSetMacro(SheppLoganCutFrequency, double);
 
+  /** Set/Get the previous kernel update size */
+  itkGetConstMacro(PreviousKernelUpdateSize, SizeType);
+  itkSetMacro(PreviousKernelUpdateSize, SizeType);
+
 protected:
   FFTRampImageFilter();
   ~FFTRampImageFilter() override = default;
--- a/include/rtkFFTVarianceRampImageFilter.h
+++ b/include/rtkFFTVarianceRampImageFilter.h
@@ -1,0 +1,84 @@
+/*=========================================================================
+ *
+ *  Copyright RTK Consortium
+ *
+ *  Licensed under the Apache License, Version 2.0 (the "License");
+ *  you may not use this file except in compliance with the License.
+ *  You may obtain a copy of the License at
+ *
+ *         http://www.apache.org/licenses/LICENSE-2.0.txt
+ *
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ *
+ *=========================================================================*/
+
+#ifndef rtkFFTVarianceRampImageFilter_h
+#define rtkFFTVarianceRampImageFilter_h
+
+#include "rtkFFTRampImageFilter.h"
+#include "rtkFFTProjectionsConvolutionImageFilter.h"
+#include "itkHalfHermitianToRealInverseFFTImageFilter.h"
+
+namespace rtk
+{
+
+/** \class FFTVarianceRampImageFilter
+ * \brief Implements the variance image filter of the filtered backprojection algorithm.
+ */
+
+template<class TInputImage, class TOutputImage=TInputImage, class TFFTPrecision=double>
+class ITK_EXPORT FFTVarianceRampImageFilter :
+  public rtk::FFTRampImageFilter<TInputImage, TOutputImage, TFFTPrecision>
+{
+public:
+  typedef rtk::FFTRampImageFilter<TInputImage, TOutputImage, TFFTPrecision> Baseclass;
+
+  /** Standard class typedefs. */
+  typedef FFTVarianceRampImageFilter Self;
+  typedef rtk::FFTRampImageFilter< TInputImage,
+                                          TOutputImage,
+                                          TFFTPrecision>     Superclass;
+  typedef itk::SmartPointer<Self>                            Pointer;
+  typedef itk::SmartPointer<const Self>                      ConstPointer;
+
+  /** Some convenient typedefs. */
+  typedef typename Baseclass::InputImageType InputImageType;
+  typedef typename Baseclass::OutputImageType OutputImageType;
+  typedef typename Baseclass::FFTPrecisionType FFTPrecisionType;
+  typedef typename Baseclass::IndexType IndexType;
+  typedef typename Baseclass::SizeType SizeType;
+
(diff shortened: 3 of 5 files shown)


SimonRit:
Thanks Jannis. I would like to add a test and I can try to implement a functional one. If I want to reconstruct the variance, should I compute variance projections, then FDKWeightProjectionFilter (VarianceMode) -> FFTVarianceRampImageFilter -> FDKBackProjectionImageFilter ?

jannisdickmann:
Hi Simon. Yes that's correct. In the FFTVarianceRampImageFilter we should probably also test for the correct calculation of finterp, which should be 0.46 without Hann filter, and I can provide expected values for other Hann values. The HannY option is currently not considered and I was just printing an error message, which probably is not ideal.

include/rtkFFTVarianceRampImageFilter.hxx
```diff
@@ -0,0 +1,220 @@
...
+  using InverseIteratorType = itk::ImageRegionIteratorWithIndex<typename FFTType::InputImageType>;
+  InverseIteratorType itiK(KernelIFFT, KernelIFFT->GetLargestPossibleRegion());
+  typename FFTType::InputImageType::PixelType sumgc = 0.;
+  typename FFTType::InputImageType::PixelType sumg2 = 0.;
+  typename FFTType::InputImageType::IndexType idxshifted;
+  for(; !itiK.IsAtEnd(); ++itiK)
+  {
+    idxshifted = itiK.GetIndex();
+    if(idxshifted[0] == 0)
+      idxshifted[0] = width-1;
```
jannisdickmann: Using `width` can actually cause problems, since the inverse Fourier of the Fourier transform is not guaranteed to have the size (see https://itk.org/Doxygen/html/classitk_1_1HalfHermitianToRealInverseFFTImageFilter.html: "the input is assumed to consist of roughly half the full complex image resulting from a real-to-complex discrete Fourier transform").
One solution would be to query KernelIFFT directly for its length. The other solution may be to use `SetActualDimensionIsOdd()` on the FFT filter (https://itk.org/Doxygen/html/classitk_1_1HalfHermitianToRealInverseFFTImageFilter.html#af072dcf520dee228202a11b6e99597f5) although I would have to test how this works.
Do you have any preference?

SimonRit: No preference, no. Like you, I don't know what are the cons of using `SetActualDimensionIsOdd()`. Pick the solution you prefer!

jannisdickmann: I added a [new commit](https://github.com/SimonRit/RTK/pull/404/commits/580c0ace80c884e689bd47b5be00e0f377727275) where I determine the size of the Fourier-transformed filter again instead of relying on the `width` variable. This seems to me the safest option.

include/rtkFDKWeightProjectionFilter.hxx
```diff
@@ -60,6 +60,11 @@ FDKWeightProjectionFilter<TInputImage, TOutputImage>::BeforeThreadedGenerateData
       m_ConstantProjectionFactor[k] *= itk::Math::abs(sdd) / (2. * sid * sid);
       m_ConstantProjectionFactor[k] *= sp.GetNorm();
     }
+    if(this->m_VarianceMode)
+    {
+      // square if in variance mode
+      m_ConstantProjectionFactor[k] = m_ConstantProjectionFactor[k] * m_ConstantProjectionFactor[k];
+    }
```
SimonRit, line 67: I'm surprised you only square the constant part of the weight. Has this been tested with a divergent geometry? For simplicity, my suggestion would be to apply this filter twice instead of adding a parameters.

include/rtkFFTRampImageFilter.h
```diff
@@ -128,6 +128,10 @@ class ITK_TEMPLATE_EXPORT FFTRampImageFilter
   itkGetConstMacro(SheppLoganCutFrequency, double);
   itkSetMacro(SheppLoganCutFrequency, double);
 
+  /** Set/Get the previous kernel update size */
+  itkGetConstMacro(PreviousKernelUpdateSize, SizeType);
+  itkSetMacro(PreviousKernelUpdateSize, SizeType);
```
SimonRit, line 133: Should it be protected? Any reason to expose this parameter to the user?

include/rtkFFTVarianceRampImageFilter.h
```diff
@@ -0,0 +1,84 @@
...
+/** \class FFTVarianceRampImageFilter
+ * \brief Implements the variance image filter of the filtered backprojection algorithm.
+ */
+
+template<class TInputImage, class TOutputImage=TInputImage, class TFFTPrecision=double>
+class ITK_EXPORT FFTVarianceRampImageFilter :
+  public rtk::FFTRampImageFilter<TInputImage, TOutputImage, TFFTPrecision>
+{
+public:
+  typedef rtk::FFTRampImageFilter<TInputImage, TOutputImage, TFFTPrecision> Baseclass;
```
SimonRit, line 38: This is Superclass in ITK, should be removed

include/rtkFFTVarianceRampImageFilter.hxx
```diff
@@ -0,0 +1,221 @@
...
+    itiK.Set(itiK.Get() * itiK.Get() * finterp);
+  }
+
+  // FFT kernel
+  typename FFTType::Pointer fftK2 = FFTType::New();
+  fftK2->SetInput( ifftK->GetOutput() );
+  fftK2->SetNumberOfThreads( this->GetNumberOfThreads() );
+  fftK2->Update();
+  this->m_KernelFFT = fftK2->GetOutput();
+
```
SimonRit, line 179: This part is the only difference. Maybe we should consider putting this in a function called by the parent, which would be empty for the parent and redefined for this class? Then the reimplementation of `UpdateFFTProjectionsConvolutionKernel` can be removed as well as the getter and setter for PreviousKernelUpdateSize.

I also wonder, could this have been done after the application of the windowing and finterp kept to the original value without the numerical integration?

SimonRit (changes requested):
Sorry for the delay. I have a few questions... Can you answer and I can implement the changes myself depending on your answers? Thanks in advance.

include/rtkFDKWeightProjectionFilter.hxx
jannisdickmann: Indeed we have almost always used parallel geometries. We have also tested it with a divergent geometry and errors were few percent, but those could of course increase if the SID is not 2m as for pCT, but smaller. I like the idea of applying the filter twice. This would also avoid the special case with `m_VarianceMode`. However, there should be some note in the `rtkFFTVarianceRampImageFilter.h` that the weights filter needs to be applied twice.

include/rtkFFTRampImageFilter.h
jannisdickmann: Setting the variable `protected` is better. I don't see a reason why it needs to be exposed. It just needs to be accessible [here](https://github.com/SimonRit/RTK/pull/404/commits/96057e50dcfabf800d2a4da3c35b25053f21c2fd#diff-2e1ecba43a5b9d07ec10fb6df83db6a8a118258d3399c302cb2270b9966c8c1bR33).

include/rtkFFTVarianceRampImageFilter.hxx
jannisdickmann: For the first part, I totally agree. It's nicer to just implement the part that actually changes. Splitting `UpdateFFTProjectionsConvolutionKernel` of `rtkFFTRampImageFilter.h` into several steps would also make it more flexible for other RTK users that just want to modfiy one component of the filter.

I'm not sure if I understood the second part right, but I'll try to motivate what I did here.

To calculate `finterp` one needs access to the filter including the windowing function in real space. `finterp` is then an integral that depends on the filters `g^2` and `g_C`, which in turn depend on the windowed filter in real space. The implementation that I propose is computationally most expensive, but also most flexible, since it can take any windowed ramp filter. Alternatively I could fit an function `finterp(Hann)` and avoid the numerical integration and going back and forth between Fourier and real space.

Right now the code does the following:
- define ramp in real space
- go to Fourier space
- apply window function
- go back to real space
- calculate `finterp`
- square filter and multiply by `finterp`
- go to Fourier space

I am not sure if the many times going back and forth between real space and Fourier space can cause problems (like the filter size change I fixed [in this commit](https://github.com/SimonRit/RTK/pull/404/commits/580c0ace80c884e689bd47b5be00e0f377727275)). Maybe it can be avoided by re-defining the squared ramp in real space?

SimonRit:
Superseded by #632.

PR #406, opened 2021-03-12
COMP: Add missing iostream include for CUDA classes

LucasGandel:
@SimonRit As discussed this seems to fix the dashboard issues, but then I get a redefinition error.


SimonRit:
Thanks. Actually, replacing `std::cout` with `itkGenericExceptionMacro` would be preferable I think, see e.g., [here](https://github.com/SimonRit/RTK/blob/master/src/rtkCudaForwardProjectionImageFilter.cu#L233). Do you mind modifying your PR?

LucasGandel:
Sure! Just did the change. After solving this issue I get:

error LNK2005: "public: virtual __cdecl itk::ExceptionObject::~ExceptionObject(void)" (??1ExceptionObject@itk@@UEAA@XZ) already defined in cuda_compile_1_generated_rtkCudaBackProjectionImageFilter.cu.obj

Let's see if dashboards reproduce it once it's merged.

PR #439, opened 2021-09-22
BUG: Fix varian reader memory bugs

### LINKED ISSUE #430: rtkVarianTest is failing occasionally on MacOS / clang
The error manifests with `itkTestDriver: Process exception: Segmentation fault`, see e.g. https://my.cdash.org/viewTest.php?onlyfailed&buildid=2059619. The problem also occurs on github actions now, see e.g. #429. The crash only occurs occasionally, like a memory error, at the end of this [loop in rtkHndImageIO.cxx](https://github.com/SimonRit/RTK/blob/master/src/rtkHndImageIO.cxx#L228L258) with `std::cout`s. But like any memory problem, I can't be sure this is where the problem is...

@agravgaard

agravgaard:
See #430 

For reference on misaligned access: https://stackoverflow.com/questions/44745953/correctness-of-misaligned-access-in-c

For reference on unsigned integer sizes: https://en.cppreference.com/w/cpp/language/types

Fixes #430

### DIFF
--- a/src/rtkHndImageIO.cxx
+++ b/src/rtkHndImageIO.cxx
@@ -18,7 +18,6 @@
 
 // std include
 #include <cstdio>
-#include <valarray>
 #include <numeric>
 
 #include "rtkHndImageIO.h"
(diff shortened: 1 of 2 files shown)


agravgaard:
I think the ComponentType is not the problem, but instead that I miscalculated the `i += 4` in the loop and it's an off-by-one error trying to write an element after the array, i.e. UB.
Adding a conditional check may or may not affect performance. I've gotten results of 1.1x faster (gcc), exactly equivalent (gcc) and 1.2x slower (clang). See for yourself https://quick-bench.com/q/roJudMNw92oObliax_nMYgg75b8

agravgaard:
Now only the `xmlparse.c` from ITK is giving errors:
```
212: /Users/andreas/Projects/ITK/Modules/ThirdParty/Expat/src/expat/xmlparse.c:4218:12: runtime error: left shift of 3590950230741598797 by 5 places cannot be represented in type 'unsigned long'
212: SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior /Users/andreas/Projects/ITK/Modules/ThirdParty/Expat/src/expat/xmlparse.c:4218:12 in
212: /Users/andreas/Projects/ITK/Modules/ThirdParty/Expat/src/expat/xmlparse.c:4218:18: runtime error: unsigned integer overflow: 13168588911346387264 + 18281801724885702730 cannot be represented in type 'unsigned long'
212: SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior /Users/andreas/Projects/ITK/Modules/ThirdParty/Expat/src/expat/xmlparse.c:4218:18 in
212:
212: Error per Pixel = 0
212: MSE = 0
212: PSNR = infdB
212: QI = 1
212:
212: Error per Pixel = 0
212: MSE = 0
212: PSNR = infdB
212: QI = 1
212:
212: Error per Pixel = 0
212: MSE = 0
212: PSNR = infdB
212: QI = 1
212:
212:
212: Test PASSED!
1/1 Test #212: rtkVarianTest ....................   Passed    3.47 sec
```

SimonRit:
Thanks. I tested it on my mac and it seems to have fixed the bug. Can you squash all the commits in one and prefix the message with "BUG:"? It would be nice to add "Fixes #430" after a blank line to have the issue automatically closed.

SimonRit:
Thanks @agravgaard !

PR #442, opened 2021-10-04
STYLE: Replace itkStaticConstMacro with static constexpr

SimonRit:
Required following InsightSoftwareConsortium/ITK#2767
Done with the following bash scripts:
for i in include/*
do
    sed -i "s/itkStaticConstMacro(\(.*\), \(.*\),
    \(.*\));/static constexpr \2 \1 = \3;/g" $i
done


SimonRit:
Merging, problems with python packages are independent of this PR

PR #443, opened 2021-10-04
COMP: add Linux x64 self-hosted CI for CUDA

SimonRit:


.github/workflows/build-test-package.yml
```diff
@@ -128,154 +133,3 @@ jobs:
         call "C:\Program Files (x86)\Microsoft Visual Studio\2019\Enterprise\VC\Auxiliary\Build\vcvars64.bat"
         ctest --output-on-failure -j 2 -V -S dashboard.cmake
```
LucasGandel: @SimonRit Can you try the following to turn extra-verbosity on? We need more info to understand which part failed.

ctest --output-on-failure -j 2 -**VV** -S dashboard.cmake

SimonRit:
@LucasGandel @Thibault-Pelletier I think this is now ready to be merged (python packages errors will be fixed when ITK 5.3rc02 will be released). The following did the tricks:
- modify `.env` and `.path` files of the self-hosted runner to set-up the environment
- use a more recent git version
- update cmake to avoid strange ninja errors (see https://my.cdash.org/build/2079229)

Please have a look and let me know what you think. I don't know if there is a better way to cleanup the self-hosted (I'm not sure that this solution is cross-platform),  I'll now try to setup a Windows runner to find out. And I would have prefer to keep the cmake version similar to the one in [ITKModuleTemplate](https://github.com/InsightSoftwareConsortium/ITKModuleTemplate).

LucasGandel:
@SimonRit The change looks good to me. Please find my comments below regarding open points:
- Clean: I confirm you have to clean selfhosted builds yourself. According to [this ](https://docs.github.com/en/actions/learn-github-actions/workflow-syntax-for-github-actions#using-a-specific-shell) documentation, using bash as shell should also work on Windows. However it looks like you are only cleaning the ITK source and build folders. My understanding is that each RTK build run on the selfhosted should be cleaned too. Do you see the number of temp folders increasing on the machine for each new build?
- CMake: To me it looks like a polluted build directory where the cache still points to the ninja version downloaded from previous builds (_temp/-445951015). Cleaning the RTK build directory should help. Do you confirm you don't clean the RTK build directory between runs?

SimonRit:
I wasn't cleaning up RTK builds indeed. Removing RTK's build directory fixed the issue with cmake 3.18. But there was one warning left https://my.cdash.org/build/2079373/configure related to env variable CUDAARCHS being read since cmake 3.20. I have added a change to set 5.2 as the default cuda architecture with RTK. I think it's ready to be merged but I'll wait for our approval @LucasGandel . Thanks in advance!

itk-module-init.cmake
```diff
@@ -3,12 +3,23 @@
 #
 list(APPEND CMAKE_MODULE_PATH ${CMAKE_CURRENT_LIST_DIR}/cmake)
 
+# Set default Cuda architecture if not provided. The first case is for backward
+# compatibility with cmake versions before 3.20 which did not handle CUDAARCHS
+# environment variable.
+if(NOT DEFINED CMAKE_CUDA_ARCHITECTURES)
+  if(NOT "$ENV{CUDAARCHS}" STREQUAL "")
```
LucasGandel: FYI Using `if(DEFINED ENV{CUDAARCHS})` might be more robust. I remember having strange behavior with such tests, but I'm fine with it if it works.

LucasGandel (approved):

LucasGandel:
LGTM! +1

itk-module-init.cmake
SimonRit: Thanks for the suggestion, I have changed it to this.

PR #451, opened 2021-10-28
STYLE: update clang-format from ITK

SimonRit:


SimonRit:
Updating style according to ITK results in failing clang-format action. Any idea @LucasGandel on how to update the action according to the latest .clang-format file in ITK?

PR #453, opened 2021-11-02
ENH: upgrade CI for python packages to ITK v5.3rc02

SimonRit:


SimonRit:
ITKv5.3rc03 is out, giving up on this one

PR #456, opened 2021-11-19
BUG: CudaFDKConeBeamReconstructionFilter not using GPU default subset size

MrTzschr:
Fixes CudaFDKConeBeamReconstructionFilter not using the default number projection subsets as given by RTK_CUDA_PROJECTIONS_SLAB_SIZE (and thus SLAB_SIZE).
See discussion on [mailinglist](https://public.kitware.com/pipermail/rtk-users/2021-November/011117.html)


SimonRit:
Thanks!

Issue #459, opened 2021-12-16
Remove circular headers


SimonRit:
After the next ITK release, apply InsightSoftwareConsortium/ITK#2958

SimonRit:
Done by 4e48dc45f4872cd8c9f9f2ab2f9c2b60dd263f46

PR #460, opened 2021-12-17
remove recursive header includes

hjmjohnson:
COMP: Remove inclusion of .hxx files as headers

The ability to include either .h or .hxx files as
header files required recursively reading the
.h files twice.  The added complexity is
unnecessary, costly, and can confuse static
analysis tools that monitor header guardes (due
to reaching the maximum depth of recursion
limits for nested #ifdefs in checking).


SimonRit:
Thanks @hjmjohnson. Fixes issue #459. However, RTK is maintained to be backward compatible with the latest ITK release (see [INSTALLATION.md](https://github.com/SimonRit/RTK/blob/master/INSTALLATION.md)) so I think I' wait for the next ITK release for merging.

SimonRit:
I think we can leave it open until ITK is released, don't you think so?

hjmjohnson:
Sure.  I was doing holiday housecleaning to prune pull requests that I do not have control of.

PR #467, opened 2022-01-28
COMP: fix compilation of external project depending on RTK with Cuda

SimonRit:


SimonRit:
@LucasGandel I think this is required until CudaCommon is split to a separate repository.

SimonRit:
What do you think @LucasGandel?

LucasGandel:
@SimonRit Sorry for the late reply. Yes this should be correct as long as ITKCudaCommon is built with add_subdirectory.

+1

LucasGandel (approved):

PR #470, opened 2022-02-14
ENH: ProjectionsReader: support RawImageIO

### LINKED ISSUE #469: ProjectionsReader support RawImageIO
Thanks a lot for this nice package.
Here is a feature suggestion:

It would be useful if the class rtk::ProjectionsReader supported reading out raw data files, in the same way than itk::ImageSeriesReader does.

```
    using ImageIOType = itk::RawImageIO<unsigned short, 3>;
    using ImageType = itk::Image<unsigned short, 3>;
    ImageIOType::Pointer io = ImageIOType::New();
    io->SetDimensions(...);
    io->SetOrigin(...);
    io->SetSpacing(...);
    io->SetByteOrderToLittleEndian();
    using ReaderType = itk::ImageSeriesReader< ImageType >;
    ReaderType::Pointer reader = ReaderType::New();
    reader->SetImageIO( io );
    reader->SetFileNames( nameGenerator->GetFileNames() );
    reader->Update();
```

Currently, this cannot be done via rtk::ProjectionsReader, as it is missing the SetImageIO functionality.

ferdymercury:
CopyPaste SetImageIO function from ImageSeriesFileReader to ProjectionsReader

Fixes https://github.com/SimonRit/RTK/issues/469

### DIFF
--- a/include/rtkProjectionsReader.h
+++ b/include/rtkProjectionsReader.h
@@ -258,6 +258,16 @@ public:
   void
   GenerateOutputInformation() override;
 
+  /** Set/Get the ImageIO helper class. Often this is created via the object
+   * factory mechanism that determines whether a particular ImageIO can
+   * read a certain file. This method provides a way to get the ImageIO
+   * instance that is created. Or you can directly specify the ImageIO
+   * to use to read a particular file in case the factory mechanism will
+   * not work properly (e.g., unknown or unusual extension). */
+  void
+  SetImageIO(itk::ImageIOBase * imageIO);
+  // itkGetModifiableObjectMacro(ImageIO, itk::ImageIOBase); this is somehow not working
+
 protected:
   ProjectionsReader();
   ~ProjectionsReader() override = default;
@@ -312,6 +322,9 @@ private:
 
   /** Image IO object which is stored to create the pipe only when required */
   itk::ImageIOBase::Pointer m_ImageIO{ nullptr };
+
+  bool m_UserSpecifiedImageIO; // keep track whether the
+                               // ImageIO is user specified
 
   /** Copy of parameters for the mini-pipeline. Parameters are checked and
    * propagated when required in the GenerateOutputInformation. Refer to the
--- a/include/rtkProjectionsReader.hxx
+++ b/include/rtkProjectionsReader.hxx
@@ -120,6 +120,9 @@ ProjectionsReader<TOutputImage>::ProjectionsReader()
   m_UpperBoundaryCropSize.Fill(0);
   m_ShrinkFactors.Fill(1);
   m_MedianRadius.Fill(0);
+
+  m_ImageIO = nullptr;
+  m_UserSpecifiedImageIO = false;
 }
 
 //--------------------------------------------------------------------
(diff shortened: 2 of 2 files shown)


ferdymercury:
With the reproducer I sent you per e-mail, I still get one error, and I am unsure what the best way would be to solve it.

`ITK ERROR: Can not use I0 in LUTbasedVariableI0RawToAttenuationImageFilter with this input (not implemented)`

Also, the input images are unsigned short, but if the logarithm is calculated, the output should be float rather than ints, right? Can the output of ProjReader be of a different type than the input?

SimonRit:
Yes, the input is often different from the output which is always float or double. Input should be integers for using the I0 option.
The ITK error here is caused by the current test [here](https://github.com/SimonRit/RTK/pull/470/files#diff-df29114932beecd7236503c98a20b751da3293158c12332be9a45c66adf5d362R172) that sets everything up the first time. It never enters this block with your current modification...
Should I try to find a solution or you'll handle it?

ferdymercury:
> Should I try to find a solution or you'll handle it?

Thanks for the reply. I can try to find a solution, however I will need some guidance, as it is my first time using RTK. If it is less work for you to implement yourself than to guide me, then please feel free to do it yourself, much appreciated.

Here my questions:

```
imageIO->SetFileName(m_FileNames[0].c_str());
imageIO->ReadImageInformation();
```
I believe these two lines above won't be needed for RawImageIO, right ?

And what about the name of the class, is it called "RawImageIO" instead of "EdfImageIO"?, which means that I have to create a similar if-else block below?

```
if ((!strcmp(imageIO->GetNameOfClass(), "EdfImageIO") &&
         imageIO->GetComponentType() == itk::ImageIOBase::IOComponentEnum::USHORT) ||
        !strcmp(imageIO->GetNameOfClass(), "XRadImageIO"))
    {

```

Is the ChangeInfoFilter necessary also for RawImageIO ? What is it doing?

Do I need to create a "RawToAttenuation" Filter analogously to "EdfRawToAttenuation" ? In what file?

```
using RawFilterType = rtk::EdfRawToAttenuationImageFilter<InputImageType, OutputImageType>;
typename RawFilterType::Pointer rawFilter = RawFilterType::New();
m_RawToAttenuationFilter = rawFilter;
```

Thanks in advance.

SimonRit:
> I believe these two lines above won't be needed for RawImageIO, right ?

I'm actually not sure but I guess not since there is no information to read, all is set manually.

> And what about the name of the class, is it called "RawImageIO" instead of "EdfImageIO"?, which means that I have to create a similar if-else block below?

No need to handle this. Since your pixel type is unsigned short, you want to go in [this block](https://github.com/SimonRit/RTK/blob/08fd9ad16fda710e1e3153ebb9a3ae37cf6f2578/include/rtkProjectionsReader.hxx#L288). So you only need to address the fact that m_ImageIO always equals imageIO, maybe using the boolean you have defined.

> Do I need to create a "RawToAttenuation" Filter analogously to "EdfRawToAttenuation" ? In what file?

No, nothing else to develop. If I remove the test `if (m_ImageIO != imageIO)`, everything already works but it's better to make sure that you enter this block once only.

ferdymercury:
Thanks Simon, I did as you suggested (see latest commit).

Now I do not get any error. However, my reconstructed image is completely empty (all zeroes).

I tried also calling ReadImageInformation (removing the if-check), but result is the same. So I am a bit lost again. I am not sure whether the Spacing and Origin of RawImageIO are being forwarded correctly to the ProjectionsReader class.

ferdymercury:
Ok, changing the output ImageType to float makes things better, now it is not all zero. Closing in favor of #472. Thanks!

PR #474, opened 2022-02-17
COMP: set CMake vars for ITKCudaCommon targets

SimonRit:
This mimicks what is done in ITKModuleEnablement.cmake for enabled
modules in addition to including the module directory.


SimonRit:
@LucasGandel to fix the [dashboard](https://my.cdash.org/index.php?project=RTK) problems

SimonRit:
Merging since only self-hosted passes and the only one with Cuda capability

PR #479, opened 2022-03-01
COMP: fix incorrect casts from scalar or vector to points

SimonRit:
Compilation errors were purposely introduced by
InsightSoftwareConsortium/ITK#3237

### DIFF
--- a/src/rtkOraGeometryReader.cxx
+++ b/src/rtkOraGeometryReader.cxx
@@ -119,7 +119,7 @@ OraGeometryReader::GenerateData()
       MetaDataDoubleType * zvecMeta =
         dynamic_cast<MetaDataDoubleType *>(dic["zdistancebaseunitcs2imagingcs_cm"].GetPointer());
       double zvec = zvecMeta->GetMetaDataObjectValue();
-      tiltTransform->SetCenter(itk::MakeVector(0., -10. * yvec, -10. * zvec));
+      tiltTransform->SetCenter(itk::MakePoint(0., -10. * yvec, -10. * zvec));
 
       sp = tiltTransform->TransformPoint(sp);
       dp = tiltTransform->TransformPoint(dp);
--- a/test/rtkoratest.cxx
+++ b/test/rtkoratest.cxx
@@ -82,8 +82,7 @@ main(int argc, char * argv[])
   spacing[1] = 2.;
   spacing[2] = 1.;
   reader->SetSpacing(spacing);
-  ReaderType::OutputImagePointType origin(0.);
-  reader->SetOrigin(0.);
+  reader->SetOrigin(itk::MakePoint(0., 0., 0.));
   ReaderType::OutputImageDirectionType direction;
   direction.SetIdentity();
   reader->SetDirection(direction);


SimonRit:
FYI @N-Dekker!

src/rtkOraGeometryReader.cxx
```diff
@@ -119,7 +119,7 @@ OraGeometryReader::GenerateData()
       MetaDataDoubleType * zvecMeta =
         dynamic_cast<MetaDataDoubleType *>(dic["zdistancebaseunitcs2imagingcs_cm"].GetPointer());
       double zvec = zvecMeta->GetMetaDataObjectValue();
-      tiltTransform->SetCenter(itk::MakeVector(0., -10. * yvec, -10. * zvec));
+      tiltTransform->SetCenter(itk::MakePoint(0., -10. * yvec, -10. * zvec));
```
N-Dekker, line 122: Thank you @SimonRit I hope you agree that https://github.com/InsightSoftwareConsortium/ITK/pull/3237, declaring those `Point(v)` constructors `explicit` does really help to make ITK user code better. 😃  In this particular case, calling `itk::MakePoint` certainly looks more appropriate than calling `itk::MakeVector`  :+1:

test/rtkoratest.cxx
```diff
@@ -82,8 +82,7 @@ main(int argc, char * argv[])
   spacing[1] = 2.;
   spacing[2] = 1.;
   reader->SetSpacing(spacing);
-  ReaderType::OutputImagePointType origin(0.);
```
N-Dekker, line 85: @SimonRit This particular line of code still compiles with `ITK_LEGACY_REMOVE=ON`, right? The intention of https://github.com/InsightSoftwareConsortium/ITK/pull/3237 was to avoid _implicit_ conversion from a single value to a `Point` (as well as from a `Vector` to a `Point`). It should still allow an initialization of the form `PointType origin(0.)`.

Of course, if the line of code isn't useful anymore, it's fine to me to have it removed anyway. 👍

SimonRit: Yes, line below was failing but not this one.

src/rtkOraGeometryReader.cxx
SimonRit: Honestly, it's not obvious to me why an `itk::Point` is not an `itk::Vector`. I have often fought against itk when using vectors, matrices and points and I'm pretty sure that there are many more places where RTK code could be improved... As far as I know, this is not well documented, but I might have missed a good explanation.

N-Dekker: @SimonRit Well, I guess that itk::Point and itk::Vector are just like in https://www.scratchapixel.com/lessons/mathematics-physics-for-computer-graphics/geometry/points-vectors-and-normals (but then in an _N_-dimensional space):

> Points and Vectors
...
> Here, a point is a position in a three-dimensional space. A vector, on the other hand, usually means a direction (and some corresponding magnitude, or size) in three-dimensional space.

Right?

SimonRit: Yes but from the same source `Three-dimensional points and vectors are of course similar in that they are both represented by the aforementioned tuple notation.` Which is why I would not have used two different types.

test/rtkoratest.cxx
```diff
@@ -82,8 +82,7 @@ main(int argc, char * argv[])
...
-  reader->SetOrigin(0.);
+  reader->SetOrigin(itk::MakePoint(0., 0., 0.));
```
N-Dekker, line 85: For the record, the following should also work fine, assuming the right `PointType` definition (double 3D, I guess):

    reader->SetOrigin(PointType());

The coordinates of a point are properly zero-initialized by the expression `PointType()`.

But of course, `reader->SetOrigin(itk::MakePoint(0., 0., 0.))` is also OK! 😃

src/rtkOraGeometryReader.cxx
N-Dekker: Points and vectors are similar, but they are not the same. At least, I find it helpful to distinguish between them, by using different types for them. Even though I did not invent `itk::Point` and `itk::Vector` myself  😃

PR #481, opened 2022-03-11
COMP: avoid implicit conversion from scalar to itk::Vector

SimonRit:
Required with ITK_LEGACY_REMOVE=ON since InsightSoftwareConsortium/ITK#3255

### DIFF
--- a/include/rtkDisplacedDetectorImageFilter.h
+++ b/include/rtkDisplacedDetectorImageFilter.h
@@ -73,7 +73,7 @@ public:
   using OutputImageType = TOutputImage;
   static constexpr unsigned int NDimension = TInputImage::ImageDimension;
   using OutputImageRegionType = typename OutputImageType::RegionType;
-  using WeightImageType = itk::Image<typename TOutputImage::InternalPixelType, 1>;
+  using WeightImageType = itk::Image<double, 1>;
 
   using GeometryType = ThreeDCircularProjectionGeometry;
   using GeometryConstPointer = GeometryType::ConstPointer;
--- a/include/rtkDrawGeometricPhantomImageFilter.h
+++ b/include/rtkDrawGeometricPhantomImageFilter.h
@@ -72,6 +72,13 @@ public:
   itkSetMacro(PhantomScale, VectorType);
   itkGetMacro(PhantomScale, VectorType);
 
+  /** Set isotropic scaling factor. */
+  virtual void
+  SetPhantomScale(const ScalarType _arg)
+  {
+    SetPhantomScale(VectorType(_arg));
+  }
+
   /** Get / Set the spatial offset of the phantom relative to its center. The
    * default value is (0, 0, 0). */
   itkSetMacro(OriginOffset, VectorType);
--- a/include/rtkDisplacedDetectorImageFilter.hxx
+++ b/include/rtkDisplacedDetectorImageFilter.hxx
@@ -197,7 +197,7 @@ DisplacedDetectorImageFilter<TInputImage, TOutputImage>::DynamicThreadedGenerate
     // No overlap, set output region to 0
     while (!itOut.IsAtEnd())
     {
-      itOut.Set(0.);
+      itOut.Set({});
       ++itOut;
     }
     return;
--- a/include/rtkGgoFunctions.h
+++ b/include/rtkGgoFunctions.h
@@ -81,7 +81,7 @@ SetConstantImageSourceFromGgo(typename TConstantImageSourceType::Pointer source,
   source->SetSpacing(imageSpacing);
   source->SetDirection(imageDirection);
   source->SetSize(imageDimension);
-  source->SetConstant(0.);
+  source->SetConstant({});
 
   // Copy output image information from an existing file, if requested
   // Overwrites parameters given in command line, if any
--- a/include/rtkJosephBackProjectionImageFilter.h
+++ b/include/rtkJosephBackProjectionImageFilter.h
@@ -54,13 +54,13 @@ public:
     return !(*this != other);
   }
 
-  inline int
+  inline TOutput
   operator()(const double        itkNotUsed(stepLengthInVoxel),
              const TCoordRepType itkNotUsed(weight),
              const TInput *      itkNotUsed(p),
              const int           itkNotUsed(i)) const
   {
-    return 0;
+    return {};
   }
 };
 
--- a/include/rtkJosephForwardProjectionImageFilter.hxx
+++ b/include/rtkJosephForwardProjectionImageFilter.hxx
@@ -269,7 +269,7 @@ JosephForwardProjectionImageFilter<TInputImage,
     }
     else
       m_ProjectedValueAccumulation(
-        threadId, itIn->Get(), itOut.Value(), 0., pixelPosition, pixelPosition, dirVox, pixelPosition, pixelPosition);
+        threadId, itIn->Get(), itOut.Value(), {}, pixelPosition, pixelPosition, dirVox, pixelPosition, pixelPosition);
   }
   delete itIn;
 }
--- a/include/rtkProjectGeometricPhantomImageFilter.h
+++ b/include/rtkProjectGeometricPhantomImageFilter.h
@@ -79,6 +79,13 @@ public:
   itkSetMacro(PhantomScale, VectorType);
   itkGetMacro(PhantomScale, VectorType);
 
+  /** Set isotropic scaling factor. */
+  virtual void
+  SetPhantomScale(const ScalarType _arg)
+  {
+    SetPhantomScale(VectorType(_arg));
+  }
+
   /** Get / Set the spatial offset of the phantom relative to its center. The
    * default value is (0, 0, 0). */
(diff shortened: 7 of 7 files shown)


include/rtkDisplacedDetectorImageFilter.h
```diff
@@ -73,7 +73,7 @@ class ITK_TEMPLATE_EXPORT DisplacedDetectorImageFilter : public itk::InPlaceImag
   using OutputImageType = TOutputImage;
   static constexpr unsigned int NDimension = TInputImage::ImageDimension;
   using OutputImageRegionType = typename OutputImageType::RegionType;
-  using WeightImageType = itk::Image<typename TOutputImage::InternalPixelType, 1>;
+  using WeightImageType = itk::Image<double, 1>;
```
N-Dekker, line 76: @SimonRit Thanks for your explanation, by mail 😃 Looks like I'm at least able to _comment_ on your PR, without being an official collaborator of the repo!

Is this little change of `WeightImageType` related to https://github.com/InsightSoftwareConsortium/ITK/pull/3255 ? Just wondering, no problem.

N-Dekker:
Does the CI (GitHub Actions, I see) have `ITK_LEGACY_REMOVE=ON` ?  If so, I guess then you fixed all possible compile errors that came from https://github.com/InsightSoftwareConsortium/ITK/pull/3255 right?

include/rtkDisplacedDetectorImageFilter.h
SimonRit: Yes. The weights in the reconstruction of both scalar and vector images. Before InsightSoftwareConsortium/ITK#3255, the weights were computed with the same pixel type as the reconstructed image (float / double or vector of float / double). This was not a problem because each pixel of this 1D image was [filled with the same value](https://github.com/SimonRit/RTK/blob/master/include/rtkDisplacedDetectorImageFilter.hxx#L268-L273) for vectors and then [multiplied to the pixels line by line](https://github.com/SimonRit/RTK/blob/master/include/rtkDisplacedDetectorImageFilter.hxx#L308). This is no longer possible to rely on implicit conversion from scalar to vectors... but the result after this change is exactly the same.

SimonRit:
> Does the CI (GitHub Actions, I see) have `ITK_LEGACY_REMOVE=ON` ? If so, I guess then you fixed all possible compile errors that came from [InsightSoftwareConsortium/ITK#3255](https://github.com/InsightSoftwareConsortium/ITK/pull/3255) right?

I don't think so, as used to be the case with ITK. But I have it locally and on night builds. Is it the best way to initialize zero values for zero or vectors in your opinion? I don't find `{}` very readable but the alternative `NumericTraits<>::ZeroValue()` is quite verbose...

include/rtkDrawGeometricPhantomImageFilter.h
```diff
@@ -72,6 +72,13 @@ class ITK_TEMPLATE_EXPORT DrawGeometricPhantomImageFilter : public itk::InPlaceI
   itkSetMacro(PhantomScale, VectorType);
   itkGetMacro(PhantomScale, VectorType);
 
+  /** Set isotropic scaling factor. */
+  virtual void
+  SetPhantomScale(const ScalarType _arg)
+  {
+    SetPhantomScale(VectorType(_arg));
```
N-Dekker, line 79: `SetPhantomScale(VectorType(_arg))` should be fine, although we recently added a new utility function, to make it clearer that the vector is _filled_ by its argument: `itk::MakeFilled`. Which would allow you to do: `SetPhantomScale(MakeFilled<VectorType>(_arg))`.

A more fundamental question is: do you really want the user to specify a single scalar value as argument to `SetPhantomScale`, when a Vector of those scalars is needed?  Otherwise, you may leave it to the user to pass `MakeFilled<VectorType>(userScalar)`, if they really need it, of course. (Instead of adding the `SetPhantomScale(const ScalarType)` overload.

SimonRit: I'm aware of it. Two answers:
- I started doing that but found it hard to maintain the code such that it's compatible with both ITK master and the latest release. The problem is that there is no preprocessor variable to test the release candidates... I would not have had the same issue with regular releases since either the major or the minor version number is incremented. At some point, I started testing `#ifdef itkMakeFilled_h` but that was not satisfactory! To be honest, I think the spirit of release candidates for ITK is no longer there, I'm not sure why.
- Then, I also realized that in this case, the user will very rarely use the vector version, the scalar version is the conventional one and he might want from time to time to do some anisotropic rescaling of phantoms. So yes, I don't think the user should bother with MakeFilled when all he wants is to rescale his phantom and having the two versions is the best option in my opinion.

N-Dekker:
>  Is it the best way to initialize zero values for zero or vectors in your opinion? 
> I don't find `{}` very readable but the alternative `NumericTraits<>::ZeroValue()` is quite verbose...

As always with C++, there are many choices. `{}` should work fine to specify a zero-filled Vector as argument, but in the case of overloading, you may need to be more specific about the argument type. You could then use C++11 `VectorType{}`, or the old C++98  `VectorType()` syntax.

If you want to be more explicit, you could explicitly pass zero as argument to the constructor: `VectorType(0.)`. Or call the new utility function, `itk::MakeFilled<VectorType>(0.)`. Those options _might_ be slightly slower than just `{}`, `VectorType{}`, or `VectorType()`. But the performance difference is probably not noticeable at all.

Does that help?

SimonRit:
> Does that help?

Not completely. I want a syntax which works for both scalar and itk::Vector pixel types. What would you suggest then?

N-Dekker:
> I want a syntax which works for both scalar and itk::Vector pixel types. What would you suggest then?

    {}
    PixelType{}
    PixelType()
    PixelType{ 0 }
    static_cast<PixelType>(0)
    NumericTraits<PixelType>::ZeroValue()

Should all work fine for both scalar and itk::Vector pixel types. Personally I would prefer the shortest possible. Because zero (or a Vector of zeros) is so common.

N-Dekker:
Regarding my ITK pull request https://github.com/InsightSoftwareConsortium/ITK/pull/3255: This `Vector` constructor was declared `explicit` already more than seven years ago, with `ITK_FUTURE_LEGACY_REMOVE=ON`: https://github.com/InsightSoftwareConsortium/ITK/commit/a8909625948547989f97385fd0b06c9f331606eb  With different people involved: Wang Chengjia, Tim Allman, Brian Helba, and Christopher Mullins. After seven years I think it's more than reasonable to finally move it forward to `ITK_LEGACY_REMOVE=ON`. (Actually I think seven years is still way too long. But someone has to do it!)

I'm sorry that it caused you extra work.

N-Dekker:
For what it's worth: one could still write a class that behaves like `itk::Vector`, but then allowing it to be filled implicitly by a scalar, like the following "ImplicitlyFillableVector":

```
  template <typename T, unsigned int VVectorDimension = 3>
  class ImplicitlyFillableVector
    : public itk::Vector<T, VVectorDimension>
  {
  public:
    using Superclass = itk::Vector<T, VVectorDimension>;
    using typename Superclass::ValueType;

    ImplicitlyFillableVector() = default;

    /// Fills the Vector by implicit conversion from a scalar value!
    ImplicitlyFillableVector(const ValueType & scalarValue)
      : Superclass( itk::MakeFilled<Superclass>(scalarValue) )
    {}

    ImplicitlyFillableVector(const Superclass & vec)
      : Superclass(vec)
    {}
  };

```

_In general_ I believe such an "ImplicitlyFillableVector" should not be necessary, but maybe in some specific cases, like the one you encountered...? A single `SetScale(const ImplicitlyFillableVector<double> &)` function might then be called either way:

    // Implicitly fills an entire vector:
    SetScale(42.0);

    // Directly passes a vector:
    SetScale(itk::MakeVector(1.0, 2.0, 4.0));

SimonRit:
> Regarding my ITK pull request [InsightSoftwareConsortium/ITK#3255](https://github.com/InsightSoftwareConsortium/ITK/pull/3255): This `Vector` constructor was declared `explicit` already more than seven years ago, with `ITK_FUTURE_LEGACY_REMOVE=ON`: [InsightSoftwareConsortium/ITK@a890962](https://github.com/InsightSoftwareConsortium/ITK/commit/a8909625948547989f97385fd0b06c9f331606eb) With different people involved: Wang Chengjia, Tim Allman, Brian Helba, and Christopher Mullins. After seven years I think it's more than reasonable to finally move it forward to `ITK_LEGACY_REMOVE=ON`. (Actually I think seven years is still way too long. But someone has to do it!)

I have nightly builds with `ITK_FUTURE_LEGACY_REMOVE=ON` which were green so I don't understand why everything turned red (from [here](https://my.cdash.org/index.php?project=RTK&date=2022-03-09) to [here](https://my.cdash.org/index.php?project=RTK&date=2022-03-10))...




> I'm sorry that it caused you extra work.

No pb, I'm happy to improve the code!

> _In general_ I believe such an "ImplicitlyFillableVector" should not be necessary, but maybe in some specific cases, like the one you encountered...?

I'm happy with the patch as is...

PR #482, opened 2022-03-13
BUG: Fix SDD parameter calculation for parallel geometry

MichaelColonel:
The source to detector distance (SDD) parameter is forced to be equal `0.0` when it calculated by [AddProjection methods](http://www.openrtk.org/Doxygen/classrtk_1_1ThreeDCircularProjectionGeometry.html#a0fb1475ed76a28cde24fac85eae18e1e) and the SDD absolute value is less than 1.E-06.

For example:
A forward projection, in case of parallel geometry, can't be calculated if `SDD == 1.E-10;` because of checks like:

1. https://github.com/SimonRit/RTK/blob/master/src/rtkThreeDCircularProjectionGeometry.cxx#L469
2. https://github.com/SimonRit/RTK/blob/master/src/rtkThreeDCircularProjectionGeometry.cxx#L570

### DIFF
--- a/include/rtkThreeDCircularProjectionGeometry.h
+++ b/include/rtkThreeDCircularProjectionGeometry.h
@@ -115,7 +115,25 @@ public:
   /** Add projection from a projection matrix. A projection matrix is defined
    * up to a scaling factor. The function here Assumes that the input matrix
    * pMat is normalized such that pMat*(x,y,z,1)'=(u,v,1)'.
-   * This code assumes that the SourceToDetectorDistance is positive. */
+   * This code assumes that the SourceToDetectorDistance is positive.
+   *
+   * \warning For parallel geometry the matrix has a special form
+   *
+   *     | R11 R12 R13 -ProjectionOffsetX |
+   * Mp =| R21 R22 R23 -ProjectionOffsetY |
+   *     |  0   0   0           1         |
+   *
+   * where
+   *
+   * | R11 R12 R13 |
+   * | R21 R22 R23 | - cropped rotation 3x3 matrix
+   * |  0   0   0  |
+   *
+   * outOfPlaneAngle must be within (-90, 90) degrees range,
+   * otherwise the rotation angles could be miscalculated.
+   *
+   * If projection matrix is for parallel geometry, then SID is set to 1000. mm
+   */
   bool
   AddProjection(const HomogeneousProjectionMatrixType & pMat);
 
@@ -265,7 +283,6 @@ public:
   {
     return this->m_MagnificationMatrices[i];
   }
-
 
   /** Get the vector containing the collimation jaw parameters. */
   const std::vector<double> &
--- a/src/rtkThreeDCircularProjectionGeometry.cxx
+++ b/src/rtkThreeDCircularProjectionGeometry.cxx
@@ -165,7 +165,7 @@ rtk::ThreeDCircularProjectionGeometry::AddProjection(const PointType &  sourcePo
   const PointType &  S = sourcePosition;          // source pos
   const PointType &  R = detectorPosition;        // detector pos
 
-  if (itk::Math::abs(r * c) > 1e-6) // non-orthogonal row/column vectors
+  if (fabs(r * c) > 1e-6) // non-orthogonal row/column vectors
     return false;
 
   // Euler angles (ZXY convention) from detector orientation in IEC-based WCS:
@@ -222,7 +222,11 @@ rtk::ThreeDCircularProjectionGeometry::AddProjection(const PointType &  sourcePo
   double SID = n[0] * S[0] + n[1] * S[1] + n[2] * S[2];
   // SDD: distance from source to detector along detector normal
   double SDD = n[0] * (S[0] - R[0]) + n[1] * (S[1] - R[1]) + n[2] * (S[2] - R[2]);
-
+  if (itk::Math::abs(SDD) < 1.0E-06)
+  {
+    itkDebugMacro(<< "SourceDetectorDistance is less than 1.0E-06. For parallel geometry SDD is set to 0.0");
+    SDD = 0.0;
+  }
   // source offset: compute source's "in-plane" x/y shift off isocenter
   VectorType Sv;
   Sv[0] = S[0];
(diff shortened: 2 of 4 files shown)


SimonRit:
I don't think I ever added a parallel projection from a matrix. When I try to do it
```Python
>>> from itk import RTK as rtk
>>> g=rtk.ThreeDCircularProjectionGeometry.New()
>>> g.AddProjection(100,0,5)
>>> g2=rtk.ThreeDCircularProjectionGeometry.New()
>>> g2.AddProjection(g.GetMatrix(0))
/work/ITK-source/ITK/Modules/ThirdParty/VNL/src/vxl/core/vnl/algo/vnl_svd.hxx: suspicious return value (3) from SVDC
/work/ITK-source/ITK/Modules/ThirdParty/VNL/src/vxl/core/vnl/algo/vnl_svd.hxx: M is 3x3
M = [ ...
            -nan             -nan             -nan 
            -nan             -nan             -nan 
            -nan             -nan              nan  ]
WARNING: In ../../../src/rtkThreeDCircularProjectionGeometry.cxx, line 309
DataObject (0x557716809ea0): Failed to AddProjection

False
```
having a sdd actually equal to 0 does not seem to be the main problem. Do you have a test that illustrates your PR?

@lesaintjerome might be able to help, he coded `AddProjection(matrix)`

MichaelColonel:
> I don't think I ever added a parallel projection from a matrix. When I try to do it

Apologies for misleading!
I've only tested with AddProjection method,
```
bool
rtk::ThreeDCircularProjectionGeometry::AddProjection(const PointType &  sourcePosition,
                                                     const PointType &  detectorPosition,
                                                     const VectorType & detectorRowVector,
                                                     const VectorType & detectorColumnVector)
```
and not with the matrix one.
I will try to make a couple of tests to show the behavior.

MichaelColonel:
One more thing, i've tested in C++ not python.

MichaelColonel:
Just an example to show the problem in my case.

Here is the data in DICOM: https://github.com/SlicerRt/SlicerRtData/tree/master/eclipse-8.1.20-phantom-prostate/Original

Parameters for parallel geometry SDD == SAD:
RowVector: 1.22465e-16 2.22045e-16 1.0
ColumnVector: 1.0 -1.22465e-16 -1.22465e-16
SourcePosition: 1.014 -990.585 -10.1067
DetectorPosition: -248.986 -990.585 -260.107

DRR for parallel geometry
Screenshot


Parameters for cone beam geometry SDD > SAD:
RowVector: 1.22465e-16 2.22045e-16 1.0
ColumnVector: 1.0 -1.22465e-16 -1.22465e-16
SourcePosition: 1.014 -990.585 -10.1067
DetectorPosition: -248.986 309.415 -260.107

DRR for cone beam geometry
Screenshot_1

MichaelColonel:
> having a sdd actually equal to 0 does not seem to be the main problem. Do you have a test that illustrates your PR?
> 
> @lesaintjerome might be able to help, he coded `AddProjection(matrix)`

You are correct.
The main problem is that for parallel geometry [the matrix](https://github.com/SimonRit/RTK/blob/master/src/rtkThreeDCircularProjectionGeometry.cxx#L251) is singular, [the determinant of the matrix](https://github.com/SimonRit/RTK/blob/master/src/rtkThreeDCircularProjectionGeometry.cxx#L264) is zero, so proper projection parameters can't be calculated.

SimonRit:
In this case, maybe we can detect it, SDD can be set to 0 and one can directly find the rotation matrix? The problem (I think) is that the sid cannot be deduced from the matrix.

MichaelColonel:
Dear @SimonRit, could you run test workflows?

The possible fix for parallel geometry is to modify projection matrix for inner usage.
Projection matrix could store:
1) Rotation matrix 3x3.
2) Projection offset X: Matrix (0, 3)
3) Projection offset Y: Matrix (1, 3)
4) SID: Matrix(2, 3)

Rotation matrix determinant is unity, this feature could be used as a flag to calculate rotation angles for parallel geometry.
In that case `AddProjection(matrix)` method can correctly setup projection parameters. IMHO.

MichaelColonel:
> Do you have a test that illustrates your PR?

I've added a test for parallel geometry in the PR. This is just my understanding of the issue.

SimonRit:
Thanks. I’m away from any computer until next week, I'll review it then.

include/rtkThreeDCircularProjectionGeometry.h
```diff
@@ -266,6 +266,20 @@ class RTK_EXPORT ThreeDCircularProjectionGeometry : public ProjectionGeometry<3>
...
+   * M =| R21 R22 R23 -ProjectionOffsetY |
+   *    | R31 R32 R33        SID         |
+   *
+   *
+   * | R11 R12 R13 |
+   * | R21 R22 R23 | - rotation 3x3 matrix (determinant == unity)
+   * | R31 R32 R33 |
+   */
+  MatrixType
+  GetMatrix(const unsigned int i) const override;
```
lesaintjerome: I am not sure I completely agree with that. IMO, the last line should be [0,0,0,1] according to [the geometry description ](http://www.openrtk.org/Doxygen/DocGeo3D.html) and the implementation of ComputeMagnificationMatrix with (sdd==0.).

src/rtkThreeDCircularProjectionGeometry.cxx
```diff
@@ -264,6 +288,38 @@ rtk::ThreeDCircularProjectionGeometry::AddProjection(const HomogeneousProjection
   double d = pMat(0, 0) * pMat(1, 1) * pMat(2, 2) + pMat(0, 1) * pMat(1, 2) * pMat(2, 0) +
              pMat(0, 2) * pMat(1, 0) * pMat(2, 1) - pMat(0, 0) * pMat(1, 2) * pMat(2, 1) -
              pMat(0, 1) * pMat(1, 0) * pMat(2, 2) - pMat(0, 2) * pMat(1, 1) * pMat(2, 0);
+
+  if ((itk::Math::abs(d) - 1.) < 1.E-06) // D(A) == 1, possible parallel geometry
```
lesaintjerome: The determinant of the matrix should be zero. And this detects the parallel geometry case.
OK for ProjectionOffsetsX and Y. The rotation matrix (which I denote R) can be inferred from the 3x3 submatrix (which I denote A):

First row of A = first row of R
Second row of A = second row of R
Third row of R is the cross-product of the first and second rows of A (i.e. of R).
From R, the angles can be computed.

include/rtkThreeDCircularProjectionGeometry.h
MichaelColonel: I quite agree that my solution is a dirty hack, but the[ projection matrix has size 3x4](http://www.openrtk.org/Doxygen/DocGeo3D.html#ParG), and the last row [0, 0, 0, 1]. So rotation matrix has, for parallel geometry, form:
```
    | R11 R12 R13  |
    | R21 R22 R23  |
    |  0   0   0   |
```
Three rotation angles can't be calculated from singular matrix.

src/rtkThreeDCircularProjectionGeometry.cxx
MichaelColonel: I will try to implement your suggestion.

MichaelColonel:
I've reworked calculation from matrix for parallel geometry.

SimonRit:
Thanks for the contribution and the test. Looks good to me if the CI comes green.

PR #484, opened 2022-03-15
Forbild locale insensitivity

SimonRit:


SimonRit:
@lesaintjerome It should fix your issue with locale configuration

PR #486, opened 2022-03-24
Corrected Parker weights references

GabrieleBelotti:
Corrected a description of the Parker filter to match the exact literature it's based upon. In particular, Parker's paper version available online still carries an error that was later amended, as seen in classic book "Principles of computerized tomographic imaging" by Kak and Slaney.

### DIFF
--- a/include/rtkParkerShortScanImageFilter.h
+++ b/include/rtkParkerShortScanImageFilter.h
@@ -29,12 +29,10 @@ namespace rtk
 
 /** \class ParkerShortScanImageFilter
  *
- * Weighting of image projections to handle off-centered panels
- * in tomography reconstruction. Based on [Wang, Med Phys, 2002].
- *
- * Note that the filter does nothing if the panel shift is less than 10%
- * of its size. Otherwise, it does the weighting described in the publication
- * and zero pads the data on the nearest side to the center.
+ * Weighting of image projections to handle short-scans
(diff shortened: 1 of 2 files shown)


SimonRit:
I think that this class is solely based on Parker. I suggest to amend this PR by
* removing the reference to [Wesarg, Med Phys, 2002] (you could open an [issue](https://github.com/SimonRit/RTK/issues) to suggest the implementation of Wesarg)
* mentioning the correction to the denominator [here](https://github.com/SimonRit/RTK/blob/master/include/rtkParkerShortScanImageFilter.hxx#L196) (which fixes a typo in equation (12) of Parker)

GabrieleBelotti:
> I think that this class is solely based on Parker. I suggest to amend this PR by
> 
> * removing the reference to [Wesarg, Med Phys, 2002] (you could open an [issue](https://github.com/SimonRit/RTK/issues) to suggest the implementation of Wesarg)
> * mentioning the correction to the denominator [here](https://github.com/SimonRit/RTK/blob/master/include/rtkParkerShortScanImageFilter.hxx#L196) (which fixes a typo in equation (12) of Parker)

Agreed! Removed Wesarg reference and added a note on Parker's typo in header and source file at relevant line.

SimonRit:
Thanks. Can you squash the commits in one with a commit message such as "DOC: Correct Parker reference and its implementation"?

SimonRit (approved):

PR #488, opened 2022-04-11
Configure wheel name when CUDA is used

LucasGandel:
Parse CMake arguments in setup.py to extract CUDA version from
CUDAToolkit_ROOT.

WARNING: This works only on Linux when users keep the default installation
path for CUDA. Future commits should use a more robust approach that works
on Windows too.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -10,8 +10,15 @@ except ImportError:
     print('  python -m pip install scikit-build')
     sys.exit(1)
 
+# Configure wheel name if CUDA is used
+wheel_name='itk-rtk'
+# Extract cuda version from the last folder name in CUDAToolkit_ROOT path.
+for arg in sys.argv:
+  if "CUDAToolkit_ROOT" in str(arg):
+    wheel_name += ('-' + arg.rsplit('/', 1)[-1])
+
 setup(
-    name='itk-rtk',
+    name=wheel_name,
     version='2.3.0',
     author='RTK Consortium',
     author_email='rtk-users@openrtk.org',


SimonRit:
Thanks it works with the self-hosted CI. I did the following to test it:
```bash
conda create --name RTKCudaPython39 python=3.9
conda activate RTKCudaPython39
pip install itk=5.3rc3 itk-core=5.3rc3 itk-filtering=5.3rc3 itk-io=5.3rc3 itk-numerics=5.3rc3 itk-registration=5.3rc3 itk-segmentation=5.3rc3 numpy
pip install itk_rtk_cuda116-2.3.0-cp39-cp39-manylinux_2_17_x86_64.manylinux2014_x86_64.whl
LD_LIBRARY_PATH=/home/srit/Downloads/cuda116/lib64:$LD_LIBRARY_PATH python FirstCudaReconstruction.py a.mha a.xml
```
and I also checked that
```bash
pip list
```
returns
```
Package          Version
---------------- ---------
certifi          2021.10.8
itk              5.3rc3
itk-core         5.3rc3
itk-filtering    5.3rc3
itk-io           5.3rc3
itk-numerics     5.3rc3
itk-registration 5.3rc3
itk-rtk-cuda116  2.3.0
itk-segmentation 5.3rc3
numpy            1.22.3
pip              21.2.4
setuptools       58.0.4
wheel            0.37.1
```
Do you know if Pypi will automatically create the new package the next time we release?
Thanks again!

SimonRit:
Can you just add `ENH:` in front of your commit message to comply with the ITK standard?

LucasGandel:
Awesome! Thanks a lot for testing!

> Do you know if Pypi will automatically create the new package the next time we release? Thanks again!

I think Pypi use the package name, so I would expect it to create the new package automatically.

SimonRit:
I can also confirm that the Cuda packages generated by #487 (https://github.com/SimonRit/RTK/actions/runs/2157035137) work ITKv5.3rc04. Awesome!

Issue #499, opened 2022-06-27
Darker center part in reconstructed image using projections with displaced detector


theysp:
I used the following configuration of CBCT device
1. Source to detector distance 1450mm (SDD)
2. Source to rotation axis 1000mm (SID)
3. Detector of dimension 43cm x 43cm displaced about 40%, left a 3cm redundant part of the detector

I tried the iterative FDK reconstruction, set displacement filter on, I got a reconstructed image with a darker central hole.
Truncation Correction from 0.0 to 1.0 has been tried as well as various Air Threshold for the averaged scatter preprocessing and Hann Cut Frequency. The darker central hole varies from dark hole to a not so dark hole with a shining halo around it. 

Since the code of displacement has not been changed for a long time, could it be improved if I use other weighting strategy? 

Thank you very much.

update on 2022-06-29
I have tried simulation on a CT data of a phantom of chest, generated DRRs using siddon method with 40% offset geometry settings. And reconstructed it using iterative fdk reconstruction. Still the dark hole in the center appears. 

![1](https://user-images.githubusercontent.com/5262537/176355042-c37e8073-b674-4678-a3df-6da6d35f895b.jpg)

As I reviewed a lot of papers, no such obvious artifacts has been reported.

SimonRit:
I don't think the problem is in the dispaced detector weighting. You can validate this with a simulation. The problem is most likely that your input dataset is imperfect (calibration, scatter, etc.) and these imperfections are enhanced by the displaced detector weighting, a known side effect of this strategy.
I'm closing the issue but you can re-open it if you still have the issue with simulations.

theysp:
Thank you, I will try simulation. And let you know if things goes well.
Is ART reconstruction would be suffer less than FDK from the imperfection of data?

theysp:
Thank you, I will try simulation. And let you know if things goes well.
Would ART reconstruction suffer less than FDK from the imperfection of data?

SimonRit:
It is always hard to fully anticipate but I think that the problem will also be here with iterative algorithms.

theysp:
Thanks.
In another word, image post processing is always mandatory. Correct?

SimonRit:
Or adequate pre-processing of the projections

theysp:
Dear Simon,

I have tried simulation on a CT data of a phantom of chest, generated DRRs using siddon method with 40% offset geometry settings. And reconstructed it using iterative fdk reconstruction. Still the dark hole in the center appears.

[...]

PR #500, opened 2022-07-04
Update rtkIterativeFDKConeBeamReconstructionFilter.hxx

theysp:
BUG: Activate displaced detector padding in iterative FDK #499

PR #501, opened 2022-07-04
BUG: Activate displaced detector padding in iterative FDK #499

theysp:
Set displaced detector filter in iterative FDK reconstruction filter to true by default, to avoid dark hole in reconstructed image from displaced detector projections. #499


SimonRit:
Thanks! Looks good to me, I'll merge when the CI is green. FYI, you could have simply amended your commit and force pushed instead of closing #500.

theysp:
> Thanks! Looks good to me, I'll merge when the CI is green. FYI, you could have simply amended your commit and force pushed instead of closing #500.

Thanks, I may do it this way in the future.

PR #503, opened 2022-07-22
DOC: Better description for the addProjection method for a parallel geometry

MichaelColonel:
Before
image

After
image


SimonRit:
Nice thanks!

Issue #504, opened 2022-07-26
Add a 'geometry tolerance' variable for pre-compiled applications


MarkGardnerUSyd:
I create geometry files using Matlab and then run the pre-compiled rtk applications (rtkfdk.exe, rtkadmmtotalvariation.exe etc) on a windows terminal (yes I know it's weird). The use of two different operating systems causes small variations in the geometry file (), giving me errors such as:

```
File: C:\RTK-2.3.0\RTK-2.3.0\src\rtkThreeDCircularProjectionGeometryXMLFileReader.cxx
Line: 162
Description: ITK ERROR: Matrix and parameters are not consistent.
Read matrix from geometry file: 
-1237.5 -0 862.666 -160000
0 -1500 -0 0
0.48163 -0 0.876374 -1000

Computed matrix from parameters:
-1237.5 0 862.667 -160000
0 -1500 0 0
0.481631 0 0.876374 -1000
```

I believe this error is displayed when the difference between the two matrices is larger than a certain value, however this value is smaller than the computation error that occurs between different operating systems. 

If you were to add an optional variable in the compiled applications that specifies the maximum allowed difference between the two geometry files, then I could increase this value so that this error no longer appears. An example could be something like:

```
rtkfdk --geometryTolerance e-5
```

Hope this makes sense.

MichaelColonel:
It could be easily done. In example of yours the inconsistency is about 0.001 (value in first row, third column) i would say it's quite significant!

SimonRit:
Thanks for the report. Can you provide the parameter values which were used to build this projection matrix? It would be good to understand the root cause of this behavior difference between OS.

MarkGardnerUSyd:
We have a Matlab Script that creates the geometry file:

     StartAngle = -180;           %First projection angle
     EndAngle = 180;
     ProjNum = 895;
     angs = linspace(StartAngle,EndAngle,ProjNum) - 90;
     x_offset = -160;
        
     for jj = 1:numel(angs)
        G(:,jj) = [angs(jj);0;0;0;0;x_offset;0;SID;SDD];
        Pmats(:,:,jj) = gtop(G(:,jj));
     end

     WriteRTKGeometryXML(Pmats,geoFile);

where the function gtop is:


```
function [P] = gtop(g)
%Takes physically meaningful parameters g to make projection matrix P
%% the P part

[...]

PR #505, opened 2022-07-27
ENH: Added projection matrix geometry tolerance parameter in XML fie reader

MichaelColonel:
Possible solution for #504.

For example cli usage: `rtkfdk --tolerance 0.0005`

### DIFF
--- a/applications/rtkadmmtotalvariation/rtkadmmtotalvariation.cxx
+++ b/applications/rtkadmmtotalvariation/rtkadmmtotalvariation.cxx
@@ -55,7 +55,7 @@ main(int argc, char * argv[])
   rtk::SetProjectionsReaderFromGgo<projectionsReaderType, args_info_rtkadmmtotalvariation>(projectionsReader,
                                                                                            args_info);
   // Projection matrix tolerance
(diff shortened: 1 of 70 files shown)


SimonRit:
Thank you for the report and the proposed fixed. I am not sure this is the right solution because we are fixing the consequence of the problem, not the cause. Maybe we should try to write the XML file to avoid the loss of precision, whatever the OS?
In case we want such a fix first, I don't like too much adding such an option to all command line tools. What do you think of using an environment variable instead? Such an approach has been used for Cuda, see https://github.com/SimonRit/RTK/blob/master/utilities/ITKCudaCommon/src/itkCudaDataManager.cxx#L46-L53

MichaelColonel:
> What do you think of using an environment variable instead?

Agreed. That's more elegant way. Environment variable can be read in XML file reader itself rather than in every command line application.

The output file stream precision is equal to 15 in XML file writer, it's difficult to understand what was the cause of such data loss if the XML writer had been used.

SimonRit:
Thanks for your work @MichaelColonel but #504 has been fixed and I'm closing this PR since it does not seem to be relevant anymore.

PR #523, opened 2022-12-09
ENH: Change default value of OSEM regularization to 0

arobert01:


SimonRit:
Do you now think it's better like this?

arobert01:
Yes, on second thought, I think it's better to have the "classical" behavior of the MLEM/OSEM algorithm by default. Furthermore, the weight of the regularization depend of the distribution that one want to reconstruct so it would be better to let the user choose this value if needed.

SimonRit:
One test is now failing so the regularization parameter should be set in it?
https://my.cdash.org/index.php?project=RTK

arobert01:
Sorry I didn't see the fail. Yes we could add the regularization parameter in the test or change a little bit the threshold of the MSE for the test (the MSE is 451.23 instead of 450). What do you think is better ?

SimonRit:
No preference.

PR #492, opened 2022-05-30
ENH: Remove ITKCudaCommon from the source tree

AlexyPellegrini:
Now builds with the remote module CudaCommon of ITK. Builds both as a remote module of ITK and as a standalone library.
This MR replaces https://github.com/SimonRit/RTK/pull/427

PR #528, opened 2023-01-27
External cuda common

SimonRit:
Supersedes and closes #492


SimonRit:
I think this is ready to be merged. Now that the CudaCommon packages have been uploaded to [pypi](https://pypi.org/project/itk-cudacommon-cuda116/), I tested the Python 3.9 version on Linux and Windows and it seems to work (FirstCudaReconstruction.py runs). Anything else I should test or do before merging?

SimonRit:
cc @AlexyPellegrini @LucasGandel

PR #530, opened 2023-01-31
COMP: Upgrade GitHub remote action to avoid MacOS warnings

SimonRit:
See report and comment in InsightSoftwareConsortium/ITK#3894


SimonRit:
The CI affected by this change is now green. Merging...

PR #532, opened 2023-02-07
BUG: Fix the inconsistency between CUDA and multi-threads(MT) code for negative SourceToDetectorDistance

theysp:
I have found that for geometry that have negative _**sdd**_ and **_sid_**, which means the projection geometry is described in a right handed coordinate system, the **_DisplacedDetectorFilter_** implemented on CUDA behaves differently, and the CUDA version obviously takes the wrong direction of the detector row and gives wrong weights on the padded projections.
After looking into the implementation, I have found that the function **_ToUntiltedCoordinateAtIsocenter_**  has been implemented differently in CUDA and MT versions.
I have altered function **_ToUntiltedCoordinateAtIsocenter_** in rtkCudaDisplacedDetector.cu, to make it have consistent results with rtkDisplacedDetector.hxx.

### DIFF
--- a/src/rtkCudaDisplacedDetectorImageFilter.cu
+++ b/src/rtkCudaDisplacedDetectorImageFilter.cu
@@ -41,7 +41,7 @@ ToUntiltedCoordinateAtIsocenter(float tiltedCoord, float sdd, float sid, float s
   const float cosa = sx / sidu;
   // the following relation refers to a note by R. Clackdoyle, title
   // "Samping a tilted detector"
-  return l * sid / (sidu - l * cosa);
+  return l * abs(sid) / (sidu - l * cosa);
 }
 
 __global__ void


theysp:
It seems the RTK is fully tested with left handed geometry, which makes the reconstructed images having opposite Z direction with the scanned objects. But right handed geometry is more nature. This issue makes me confused all the time. Do you know why left handed geometry and clock-wise rotation angle being used in a lot of CBCT systems?@SimonRit

SimonRit:
Thanks for your contribution. Can you please update the commit message to follow [ITK guidelines](https://github.com/InsightSoftwareConsortium/ITK/blob/master/CONTRIBUTING.md)?

The inconsistency has been introduced by ad58d3c110b98254e9ff33e4de83697e852baaf6. Maybe other Cuda filters need to be updated too. I would appreciate it if you have the time to check.

I'm not sure I understand what you mean by left handed... but I usually have positive source-to-isocenter and source-to-detector distances with the systems I deal with.

theysp:
Thank you for the clue. 
I have checked the revisions you've done in [ad58d3c](https://github.com/RTKConsortium/RTK/commit/ad58d3c110b98254e9ff33e4de83697e852baaf6). 
This commit is OK with consistent between the CUDA and non-CUDA implementations.

The inconsistency happens between the rtkDisplacedDetectorImageFilter and rtkCudaDisplacedDetectorImageFilter. I think this error exists because displaced detector images is not used a lot with negative SDD as you said.



> Thanks for your contribution. Can you please update the commit message to follow [ITK guidelines](https://github.com/InsightSoftwareConsortium/ITK/blob/master/CONTRIBUTING.md)?
> 
> The inconsistency has been introduced by [ad58d3c](https://github.com/RTKConsortium/RTK/commit/ad58d3c110b98254e9ff33e4de83697e852baaf6). Maybe other Cuda filters need to be updated too. I would appreciate it if you have the time to check.
> 
> I'm not sure I understand what you mean by left handed... but I usually have positive source-to-isocenter and source-to-detector distances with the systems I deal with.

SimonRit:
I see you've updated the PR title, can you amend the commit message instead? The prefix `BUG:` is good but the message should not be longer than 72 characters, maybe
`BUG: Fix CUDA/CPU inconsistency in DisplacedDetectorImageFilter`
Then you can explain the negative "distance" in the message body.
If you don't mind, can you also add a line in [rtkdisplaceddetectorcompcudatest.cxx](https://github.com/RTKConsortium/RTK/blob/master/test/rtkdisplaceddetectorcompcudatest.cxx#L53-L56) to test a negative SID?
Thanks again!

theysp:
> I see you've updated the PR title, can you amend the commit message instead? The prefix `BUG:` is good but the message should not be longer than 72 characters, maybe `BUG: Fix CUDA/CPU inconsistency in DisplacedDetectorImageFilter` Then you can explain the negative "distance" in the message body. If you don't mind, can you also add a line in [rtkdisplaceddetectorcompcudatest.cxx](https://github.com/RTKConsortium/RTK/blob/master/test/rtkdisplaceddetectorcompcudatest.cxx#L53-L56) to test a negative SID? Thanks again!

Currently I am in a environment that's a little difficult to have some test.. I will add a new line in [rtkdisplaceddetectorcompcudatest.cxx](https://github.com/RTKConsortium/RTK/blob/master/test/rtkdisplaceddetectorcompcudatest.cxx#L53-L56) to test a negative SID later in a week.

PR #543, opened 2023-04-18
ENH: Add superior and inferior clip images for Joseph-based forward projector

arobert01:
The superior and inferior clip images define for each pixel of the projections the value of the superior and inferior clip of the ray emitted from that pixel. Each ray is clipped from
source+inferior_clip*(pixel-source) to
source+superior_clip*(pixel-source) with inferior_clip and superior_clip equal 0 and 1 by default.

### DIFF
--- a/applications/rtkforwardprojections/rtkforwardprojections.ggo
+++ b/applications/rtkforwardprojections/rtkforwardprojections.ggo
@@ -13,6 +13,6 @@ section "Projectors"
 option "fp"    f "Forward projection method" values="Joseph","JosephAttenuated","CudaRayCast","Zeng" enum no default="Joseph"
 option "attenuationmap" - "Attenuation map relative to the volume to perfom the attenuation correction"   string  no
 option "sigmazero" - "PSF value at a distance of 0 meter of the detector"   double  no
-option "alphapsf" - "Slope of the PSF against the detector distance"   double  no 
+option "alphapsf" - "Slope of the PSF against the detector distance"   double  no
 option "inferiorclipimage" - "Value of the inferior clip of the ray for each pixel of the projections (only with Joseph-based projector)" string no
 option "superiorclipimage" - "Value of the superior clip of the ray for each pixel of the projections (only with Joseph-based projector)" string no
--- a/applications/rtkforwardprojections/rtkforwardprojections.ggo
+++ b/applications/rtkforwardprojections/rtkforwardprojections.ggo
@@ -14,3 +14,5 @@ option "fp"    f "Forward projection method" values="Joseph","JosephAttenuated",
 option "attenuationmap" - "Attenuation map relative to the volume to perfom the attenuation correction"   string  no
 option "sigmazero" - "PSF value at a distance of 0 meter of the detector"   double  no
 option "alphapsf" - "Slope of the PSF against the detector distance"   double  no 
+option "inferiorclipimage" - "Value of the inferior clip of the ray for each pixel of the projections (only with Joseph-based projector)" string no
+option "superiorclipimage" - "Value of the superior clip of the ray for each pixel of the projections (only with Joseph-based projector)" string no
--- a/include/rtkJosephForwardAttenuatedProjectionImageFilter.hxx
+++ b/include/rtkJosephForwardAttenuatedProjectionImageFilter.hxx
@@ -81,6 +81,7 @@ JosephForwardAttenuatedProjectionImageFilter<TInputImage,
                                              TProjectedValueAccumulation,
                                              TComputeAttenuationCorrection>::VerifyInputInformation() const
 {
+  Superclass::VerifyInputInformation();
   using ImageBaseType = const itk::ImageBase<InputImageDimension>;
 
   ImageBaseType * inputPtr1 = nullptr;
@@ -93,7 +94,7 @@ JosephForwardAttenuatedProjectionImageFilter<TInputImage,
     // method since it returns the input as a pointer to a
     // DataObject as opposed to the subclass version which
     // static_casts the input to an TInputImage).
-    if (it.GetName() != "Primary")
+    if (it.GetName() == "_1")
     {
       inputPtr1 = dynamic_cast<ImageBaseType *>(it.GetInput());
     }
@@ -105,7 +106,7 @@ JosephForwardAttenuatedProjectionImageFilter<TInputImage,
 
   for (; !it.IsAtEnd(); ++it)
   {
-    if (it.GetName() != "Primary")
+    if (it.GetName() == "_2")
     {
       auto * inputPtrN = dynamic_cast<ImageBaseType *>(it.GetInput());
       // Physical space computation only matters if we're using two
--- a/include/rtkJosephForwardProjectionImageFilter.h
+++ b/include/rtkJosephForwardProjectionImageFilter.h
@@ -306,7 +306,7 @@ protected:
   ThreadedGenerateData(const OutputImageRegionType & outputRegionForThread, ThreadIdType threadId) override;
 
   /** If a third input is given, it should be in the same physical space
-   * than the first one. */
+   * as the first one. */
   void
   VerifyInputInformation() const override;
 
--- a/include/rtkJosephForwardProjectionImageFilter.h
+++ b/include/rtkJosephForwardProjectionImageFilter.h
@@ -183,6 +183,8 @@ public:
   using OutputImageRegionType = typename TOutputImage::RegionType;
   using CoordRepType = double;
   using VectorType = itk::Vector<CoordRepType, TInputImage::ImageDimension>;
+  using TClipImageType = itk::Image<double, TOutputImage::ImageDimension>;
+  using TClipImagePointer = typename TClipImageType::Pointer;
 
   /** Method for creation through the object factory. */
   itkNewMacro(Self);
@@ -254,6 +256,36 @@ public:
     }
   }
 
+  /** Set/Get the inferior clip image. Each pixel of the image
+   ** corresponds to the value of the inferior clip of the ray
+   ** emitted from that pixel. */
+  void
+  SetInferiorClipImage(const TClipImageType * inferiorClipImage)
+  {
+    // Process object is not const-correct so the const casting is required.
+    this->SetInput("InferiorClipImage", const_cast<TClipImageType *>(inferiorClipImage));
+  }
+  typename TClipImageType::ConstPointer
+  GetInferiorClipImage()
+  {
+    return static_cast<const TClipImageType *>(this->itk::ProcessObject::GetInput("InferiorClipImage"));
+  }
+
+  /** Set/Get the superior clip image. Each pixel of the image
+   ** corresponds to the value of the superior clip of the ray
+   ** emitted from that pixel. */
+  void
+  SetSuperiorClipImage(const TClipImageType * superiorClipImage)
+  {
+    // Process object is not const-correct so the const casting is required.
+    this->SetInput("SuperiorClipImage", const_cast<TClipImageType *>(superiorClipImage));
+  }
+  typename TClipImageType::ConstPointer
+  GetSuperiorClipImage()
+  {
+    return static_cast<const TClipImageType *>(this->itk::ProcessObject::GetInput("SuperiorClipImage"));
+  }
+
   /** Each ray is clipped from source+m_InferiorClip*(pixel-source) to
   ** source+m_SuperiorClip*(pixel-source) with m_InferiorClip and
   ** m_SuperiorClip equal 0 and 1 by default. */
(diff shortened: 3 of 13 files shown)


include/rtkJosephForwardProjectionImageFilter.hxx
```diff
@@ -121,16 +242,32 @@ JosephForwardProjectionImageFilter<TInputImage,
         mainDir = i;
     }
 
+    // m_InferiorClip and m_SuperiorClip are understood in the sense of a
+    // source-to-pixel vector. Since we go from pixel-to-source, we invert them.
+    double inferiorClip = 1. - m_SuperiorClip;
+    double superiorClip = 1. - m_InferiorClip;
+    if (this->GetInferiorClipImage().IsNotNull())
+    {
+      double superiorClipImage = 1 - this->GetInferiorClipImage()->GetPixel(pixelIndex);
```
SimonRit: I would use an iterator on the InferiorClipImage instead to be more efficient

include/rtkJosephForwardProjectionImageFilter.hxx
```diff
@@ -121,16 +242,32 @@ JosephForwardProjectionImageFilter<TInputImage,
...
     BoxShape::ScalarType nearDist = NAN, farDist = NAN;
-    if (box->IsIntersectedByRay(pixelPosition, dirVox, nearDist, farDist) &&
-        farDist >= 0. && // check if detector after the source
-        nearDist <= 1.)  // check if detector after or in the volume
+    bool                 isIntersectByRay = box->IsIntersectedByRay(pixelPosition, dirVox, nearDist, farDist);
+    // Clip the casting between source and pixel of the detector
+    nearDist = std::max(nearDist, inferiorClip);
+    farDist = std::min(farDist, superiorClip);
+
+    if (isIntersectByRay && farDist > 0. && // check if detector after the source
```
SimonRit, line 279: Please split in two lines for clarity

include/rtkJosephForwardProjectionImageFilter.hxx
```diff
@@ -121,16 +242,32 @@ JosephForwardProjectionImageFilter<TInputImage,
         mainDir = i;
     }
 
+    // m_InferiorClip and m_SuperiorClip are understood in the sense of a
+    // source-to-pixel vector. Since we go from pixel-to-source, we invert them.
+    double inferiorClip = 1. - m_SuperiorClip;
+    double superiorClip = 1. - m_InferiorClip;
```
SimonRit: Why has this been moved in the loop? It makes more sense to keep it out of the loop

include/rtkJosephForwardProjectionImageFilter.hxx
```diff
@@ -45,6 +45,130 @@ JosephForwardProjectionImageFilter<TInputImage,
...
+            directionString.setf(std::ios::scientific);
+            directionString.precision(7);
+            directionString << "InputImage Direction: " << inputPtr1->GetDirection() << ", InputImage" << it.GetName()
+                            << " Direction: " << inputPtrN->GetDirection() << std::endl;
+            directionString << "\tTolerance: " << Self::GetGlobalDefaultDirectionTolerance() << std::endl;
+          }
+          itkExceptionMacro(<< "Inputs do not occupy the same physical space! " << std::endl
+                            << originString.str() << spacingString.str() << directionString.str());
+        }
+      }
```
SimonRit, line 166: This is copied from ImageToImageFilter right? It's a pity there is not a way to avoid copy/pasting the code. Unless if you set the different inputs to 0 before calling the superclass method?

applications/rtkforwardprojections/rtkforwardprojections.ggo
```diff
@@ -14,3 +14,5 @@ option "fp"    f "Forward projection method" values="Joseph","JosephAttenuated",
 option "attenuationmap" - "Attenuation map relative to the volume to perfom the attenuation correction"   string  no
 option "sigmazero" - "PSF value at a distance of 0 meter of the detector"   double  no
 option "alphapsf" - "Slope of the PSF against the detector distance"   double  no 
+option "inferiorclipimage" - "Value of the inferior clip of the ray for each pixel of the projections (only with Joseph-based projector)" string no
+option "superiorclipimage" - "Value of the superior clip of the ray for each pixel of the projections (only with Joseph-based projector)" string no
```
SimonRit, line 18: Shouldn't this be in rtkprojectors_section.ggo?

SimonRit (changes requested):

include/rtkJosephForwardProjectionImageFilter.hxx
arobert01: Yes my bad, I moved it out of the loop

applications/rtkforwardprojections/rtkforwardprojections.ggo
arobert01: Yes you are right, it should also be in rtkprojectors_section.ggo. In this case, we should also add the inferior and superior clip images in all the applications that could use the Joseph-based projectors  (rtkosem, rtksart, etc ...). Do you want me to do it ?

arobert01: Also should I add the superior and inferior clip images in the Joseph-based backprojectors ?

SimonRit: Yes please if it's not too much work. I'm guessing it just have to be included in https://github.com/RTKConsortium/RTK/blob/master/include/rtkGgoFunctions.h?

SimonRit: > Also should I add the superior and inferior clip images in the Joseph-based backprojectors ?

That's probably more work, let's not do it yet...

include/rtkJosephForwardProjectionImageFilter.hxx
arobert01: In my understanding, in ImageToImageFilter the method verify that all the inputs are in the same physical space but in our case we just want to verify that, if given, the clip images are in the same physical space than the first input corresponding to the projections. I don't know if I can do this check by calling the superclass method.

SimonRit: My idea is the following. You have three inputs, A B C. You want A and C to have the same information, not B. You can maybe do
```
tmpB = B
B = NULL
Superclass::VerifyInputinformation()
B = tmpB
```
?
A much more pleasant implementation would be to create a function in ImageBase, e.g., `HasSameInformation(Self *img2)` in which you would move most of the `ImageToImage::VerifyInputInformation` code to reuse it, e.g., here.

applications/rtkforwardprojections/rtkforwardprojections.ggo
arobert01: > Yes please if it's not too much work. I'm guessing it just have to be included in https://github.com/RTKConsortium/RTK/blob/master/include/rtkGgoFunctions.h?

Actually, it's more work than I thought. For the reconstructions filters, like OSEM for example, the projectors are instantiate inside the method GenerateOutputInformation() of the filter : https://github.com/RTKConsortium/RTK/blob/abe956fbda210952ea3796d15e2077e6844fe1bd/include/rtkOSEMConeBeamReconstructionFilter.hxx#L82-L241 so to be able to give the clip images to the projector I think we also need to give them as input of the OSEM filter (or create setter and getter). Maybe I missing something simpler here ?

include/rtkJosephForwardProjectionImageFilter.hxx
arobert01: Thanks for the ideas. I agree than the second solution is more elegant, I can do it and create a pull request on ITK side if you want.

SimonRit: Of course I agree. Thanks!

applications/rtkforwardprojections/rtkforwardprojections.ggo
SimonRit: It would be exactly the same as what you've implemented for attenuation images. I don't remember how you did that and whether it would be a lot of work.

include/rtkJosephForwardProjectionImageFilter.hxx
arobert01: Actually it has to be in the loop since now the superior and inferior clip could be pixel dependent if clip image are given

SimonRit: Of course but you don't need to redo the calculation if there is no clip images... You can keep this out of the loop I believe.

SimonRit (changes requested):
Thanks! Closely looking at this, I think that all iterative applications can set the attenuation, PSF and clip image parameters in existing methods SetForwardProjectionFromGgo and SetBackProjectionFromGgo. This would mean moving the projectors code in rtkosem to these functions. What do you think? Should we change this in this PR or do it in a subsequent PR?

arobert01:
> Thanks! Closely looking at this, I think that all iterative applications can set the attenuation, PSF and clip image parameters in existing methods SetForwardProjectionFromGgo and SetBackProjectionFromGgo. This would mean moving the projectors code in rtkosem to these functions. What do you think? Should we change this in this PR or do it in a subsequent PR?

Yes you're right ! I can do it and amend the last commit I did.

include/rtkJosephForwardProjectionImageFilter.h
```diff
@@ -266,14 +298,17 @@ class ITK_TEMPLATE_EXPORT JosephForwardProjectionImageFilter
...
+  void
+  GenerateInputRequestedRegion() override;
+
   void
   ThreadedGenerateData(const OutputImageRegionType & outputRegionForThread, ThreadIdType threadId) override;
 
-  /** The two inputs should not be in the same space so there is nothing
-   * to verify. */
+  /** If a third input is given, it should be in the same physical space
+   * than the first one. */
```
SimonRit: than -> as

include/rtkJosephForwardAttenuatedProjectionImageFilter.hxx
```diff
@@ -93,7 +94,7 @@ JosephForwardAttenuatedProjectionImageFilter<TInputImage,
     // method since it returns the input as a pointer to a
     // DataObject as opposed to the subclass version which
     // static_casts the input to an TInputImage).
-    if (it.GetName() != "Primary")
+    if (it.GetName() == "_1")
```
SimonRit, line 97: I'm not sure what happened here. `Primary` is no longer a valid name?

include/rtkJosephForwardProjectionImageFilter.hxx
SimonRit: As discussed, will be addressed separately.

include/rtkJosephForwardProjectionImageFilter.hxx
```diff
@@ -108,8 +243,9 @@ JosephForwardProjectionImageFilter<TInputImage,
   typename BoxShape::VectorType stepMM, np, fp;
   for (unsigned int pix = 0; pix < outputRegionForThread.GetNumberOfPixels(); pix++, itIn->Next(), ++itOut)
   {
-    typename InputRegionIterator::PointType pixelPosition = itIn->GetPixelPosition();
-    typename InputRegionIterator::PointType dirVox = -itIn->GetSourceToPixel();
+    typename InputRegionIterator::PointType  pixelPosition = itIn->GetPixelPosition();
+    typename InputRegionIterator::PointType  dirVox = -itIn->GetSourceToPixel();
+    typename InputRegionIterator ::IndexType pixelIndex = itIn->GetIndex();
```
SimonRit: Is `pixelIndex` still used?

SimonRit (changes requested):
Some minor suggestions to implement rapidly. Two things for the record:
- I think the code in `SetForwardProjectionFromGgo` can be moved to `CreateForwardProjectionFromGgo` to be used in `SetForwardProjectionFromGgo` and `rtkforwardprojections.cxx`
- Maybe --geometry should be moved in `rtkprojectors_section.ggo`.
But these points can be addressed in a follow-up PR.

include/rtkJosephForwardAttenuatedProjectionImageFilter.hxx
arobert01: Yes `Primary` is a valid name but it is not the same input as `_1`.  `Primary` corresponds to the input 0 (the projections stack) and `_1` to the input 1 (the volume). I changed it because before to check that the attenuation map and the volume had the same informations we were testing that all the inputs expect the projections were in the same physical space.  But this is no longer valid because, if given, the clip images should be in the same physical space than the projections. So in this piece of code I specifically test that the volume (`_1`) has the same information than the attenuation map (`_2`).

Issue #454, opened 2021-11-10
Boellaard scatter correction does not behave as expected


SimonRit:
The issue has been reported on the mailing list
https://public.kitware.com/pipermail/rtk-users/2021-November/011108.html

GabrieleBelotti:
I digged around while testing the filter on monte carlo generated images.
1) Using the filter with parameters SPR 0.5 (as suggested by Park et al http://dx.doi.org/10.1118/1.4923179 and adherent to my average(S)/average(P) ratio) AirThreshold 32000 and NonNegativity 10 I get consistent improvements when comparing reconstructions to the original CT. MAE is ~halved across all acquisitions.
2) For now, the only dubious part of RTK implementation is https://github.com/SimonRit/RTK/blob/37b2cee717cf328bac83c1efb602f793430ab19b/include/rtkBoellaardScatterCorrectionImageFilter.hxx#L90 . As the averaging of the value "behind the patient" is conducted of the whole projection pixel count. I believe it should be averaged on the number of pixels that are contributing to the sum (those > AirThreshold). 

I'm still experimenting and I will regenerate my dataset asap by the latter change to the filter. Will let you know about this.

GabrieleBelotti:
Ok, this seems to be slightly more complicated than I anticipated:

1. The threshold was selecting Air instead of the opposite. Easily fixed by making it an upper threshold (>= to <=).
2. Dividing by the actual number of pixels under the threshold means the correction value is more likely to hit the nonnegativity constraint for the same SPR. But it should replicate Elekta software behavior better (?). 

There are two papers from Boellaard in 1997 on the same topic and both agree on the flatness of scattered dose distribution on the image plane at large air gaps (>=50cm):

1. [https://doi.org/10.1016/S0167-8140(97)00073-X](url). 
2. [https://doi.org/10.1118/1.598066](url).
I find figures 5a and 5b of the second paper to be particularly descriptive.

Nonetheless, It would be good to double-check on an Elekta database.

About the original report: Even though the user might have hit the nonnegativity constraint, I find it weird that no corrections were applied. In fact, each projection is corrected by (smallestValue - m_NonNegativityConstraintThreshold) in that case.

SimonRit:
It's a bit hard to connect my knowledge of Marcel's implementation and Boellaard's Med Phys paper... But here is how I see it. In https://doi.org/10.1118/1.598066, equation (5) states that
[...]

PR #547, opened 2023-04-27
BUG: Fixes on Boellaard Scatter Filter

GabrieleBelotti:
Changed the condition on AirThreshold value (less than AirThreshold now) and revisited the averaging of the "averageBehindPatient" value. Discussion on #454.

### DIFF
--- a/include/rtkBoellaardScatterCorrectionImageFilter.hxx
+++ b/include/rtkBoellaardScatterCorrectionImageFilter.hxx
@@ -89,7 +89,8 @@ BoellaardScatterCorrectionImageFilter<TInputImage, TOutputImage>::ThreadedGenera
       }
       ++itInSlice;
     }
-    if (npixelBehindPatient > 1) npixelBehindPatient--;
+    if (npixelBehindPatient > 1)
+      npixelBehindPatient--;
     averageBehindPatient /= npixelBehindPatient;
 
     // Compute constant correction
--- a/include/rtkBoellaardScatterCorrectionImageFilter.hxx
+++ b/include/rtkBoellaardScatterCorrectionImageFilter.hxx
@@ -77,17 +77,20 @@ BoellaardScatterCorrectionImageFilter<TInputImage, TOutputImage>::ThreadedGenera
 
     // Retrieve useful characteristics of current slice
     double averageBehindPatient = 0.;
+    unsigned int npixelBehindPatient = 1;
     double smallestValue = itk::NumericTraits<double>::max();
     for (unsigned int i = 0; i < npixelPerSlice; i++)
     {
       smallestValue = std::min(smallestValue, (double)itInSlice.Get());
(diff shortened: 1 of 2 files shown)


include/rtkBoellaardScatterCorrectionImageFilter.hxx
```diff
@@ -77,17 +77,21 @@ BoellaardScatterCorrectionImageFilter<TInputImage, TOutputImage>::ThreadedGenera
 
     // Retrieve useful characteristics of current slice
     double averageBehindPatient = 0.;
+    unsigned int npixelBehindPatient = 1;
     double smallestValue = itk::NumericTraits<double>::max();
     for (unsigned int i = 0; i < npixelPerSlice; i++)
     {
       smallestValue = std::min(smallestValue, (double)itInSlice.Get());
-      if (itInSlice.Get() >= m_AirThreshold)
+      if (itInSlice.Get() <= m_AirThreshold)
```
SimonRit, line 84: Thanks, it's a good change!

include/rtkBoellaardScatterCorrectionImageFilter.hxx
```diff
@@ -77,17 +77,21 @@ BoellaardScatterCorrectionImageFilter<TInputImage, TOutputImage>::ThreadedGenera
...
       {
         averageBehindPatient += itInSlice.Get();
+        npixelBehindPatient++;
       }
       ++itInSlice;
     }
-    averageBehindPatient /= npixelPerSlice;
+    if (npixelBehindPatient > 1)
```
SimonRit: I don't think this is correct. This new change means that if your patient is 1 pixel big or 1000000 pixels big in the projection domain, you have the same amount of scatter (considering that all pixels have the same attenuation). I need to take the time to derive the code from Boellaard's paper to make this clearer for you. More on this tonight or tomorrow...

SimonRit (changes requested):
I think the non negativity will be less often hit with the comment I made. As far as I can remember, the non negativity code is as you describe it and I don't have a better solution.

SimonRit:
I think you can modify this PR with only the change on the threshold. The rest should stay I believe and it would be good to have a few documentation notes. If you'd like, I can prepare a new PR later on.

include/rtkBoellaardScatterCorrectionImageFilter.hxx
GabrieleBelotti: Will remove it!

GabrieleBelotti:
> I think you can modify this PR with only the change on the threshold. The rest should stay I believe and it would be good to have a few documentation notes. If you'd like, I can prepare a new PR later on.

Agreed on the documentation notes. I've made the proper corrections. Let me know if I should do anything else. I believe you should be able to squash commits when merging the pull request.

SimonRit:
Fixes #454.

Issue #564, opened 2023-11-22
Step size should be configurable in iterative recon using CudaForwardProjectionImageFilter


SimonRit:
The Cuda forward projection is a ray tracing with tri-linear interpolation. The step size is configurable
https://github.com/RTKConsortium/RTK/blob/master/include/rtkCudaForwardProjectionImageFilter.h#L73
but the option has not been propagated to iterative reconstruction algorithms. This should be implemented for fine voxel reconstruction (<1 unit).
@Nickstaut

SimonRit:
Fixed by #565

PR #565, opened 2023-11-23
ENH: Allow to configure step size for iterative recon using Cuda ray-based projectors

arobert01:
This commit fixes the issue #564.


SimonRit:
Looks good, thanks. The commit message is a bit long, it should be 72 characters see [here](https://docs.itk.org/en/latest/contributing/index.html). Maybe "ENH: Step size accessors for Cuda ray tracing in iterative recon"

SimonRit:
Fixes #564.

PR #618, opened 2024-09-11
ENH: Account for lateral detector offsets in Zeng's forward projector

SimonRit:


arobert01:
Looks good to me, sorry I didn't take it into account in the first place. Should we do the same for the back projector ?

SimonRit:
> Looks good to me, sorry I didn't take it into account in the first place. Should we do the same for the back projector ?

Good point, I have force-pushed a modification. Note that taking the backprojector is not exactly the adjoint of the forward, I believe we would need to splat the rotation instead of taking the inverse. But that can be done in another PR.

include/rtkZengBackProjectionImageFilter.hxx
```diff
@@ -374,12 +376,15 @@ ZengBackProjectionImageFilter<TInputImage, TOutputImage>::GenerateData()
...
       indexSlice[Dimension - 1] = startSlice;
       m_ConstantVolumeSource->GetOutput()->TransformIndexToPhysicalPoint(indexSlice, pointSlice);
-      dist = geometry->GetSourceToIsocenterDistances()[nbProjections + indexProj] +
+      dist = geometry->GetSourceToIsocenterDistances()[nbProjections] +
              pointSlice.GetVectorFromOrigin() * m_VectorOrthogonalDetector;
     }
     if (this->GetInput(2))
     {
       m_AttenuationMapTransform->SetRotation(0., angle, 0.);
+      m_Transform->SetTranslation(itk::MakeVector(
```
arobert01: Should not it be m_AttenuationMapTransform instead of m_Transform ?

PR #620, opened 2024-09-19
ENH: Add option to compute the Cramer-Rao Lower Bound.

arobert01:
Add option to obtain the Cramer-Rao Lower Bound (CRLB) when computing the photon counts using rtkSpectralForwardModelImageFilter. The CRLB corresponds to the lower bound of the variance of any unbiased estimator. Therefore, it could be used to estimate the variance that would be obtained after material decomposition.

### DIFF
--- a/include/rtkSpectralForwardModelImageFilter.hxx
+++ b/include/rtkSpectralForwardModelImageFilter.hxx
@@ -261,6 +261,21 @@ template <typename DecomposedProjectionsType,
           typename IncidentSpectrumImageType,
           typename DetectorResponseImageType,
           typename MaterialAttenuationsImageType>
+typename MeasuredProjectionsType::ConstPointer
+SpectralForwardModelImageFilter<DecomposedProjectionsType,
+                                MeasuredProjectionsType,
+                                IncidentSpectrumImageType,
+                                DetectorResponseImageType,
+                                MaterialAttenuationsImageType>::GetOutputVariances()
+{
+  return static_cast<const MeasuredProjectionsType *>(this->GetOutput(1));
+}
+
+template <typename DecomposedProjectionsType,
+          typename MeasuredProjectionsType,
+          typename IncidentSpectrumImageType,
+          typename DetectorResponseImageType,
+          typename MaterialAttenuationsImageType>
 itk::DataObject::Pointer
 SpectralForwardModelImageFilter<DecomposedProjectionsType,
                                 MeasuredProjectionsType,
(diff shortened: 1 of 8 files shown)


include/rtkSpectralForwardModelImageFilter.hxx
```diff
@@ -470,8 +471,20 @@ SpectralForwardModelImageFilter<DecomposedProjectionsType,
...
+        photon_counts.SetData(forward.data_block());
+        cost->SetMeasuredData(photon_counts);
+        cost->ComputeFischerMatrix(in);
+        output1It.Set(cost->GetCramerRaoLowerBound());
+      }
+      else
+      {
+        output1It.Set(
+          itk::VariableLengthVector<double>(cost->GetVariances(in).data_block(), this->m_NumberOfSpectralBins));
+      }
```
SimonRit: I'm not sure I understand this, it seems to disable the previous mode which was already for `m_IsSpectralCT` `true`. Was the behavior different before?

arobert01: The function `GetVariances` was not implemented for spectral CT. It was only returning meaningless results (see [here](https://github.com/RTKConsortium/RTK/blob/master/include/rtkProjectionsDecompositionNegativeLogLikelihood.h#L236-L243)). Furthermore, it was no possible to set the boolean `m_ComputeVariances` to true with the `rtkspectralforwardmodel` application.

SimonRit: Ok... Do you understand what it does for dual energy?
https://github.com/RTKConsortium/RTK/blob/master/include/rtkDualEnergyNegativeLogLikelihood.h#L126
I have a hard time understanding this code and the [commit message](https://github.com/RTKConsortium/RTK/commit/0b3cf921f8422a2f96604786e56143e376bebadb) does not help

arobert01: If I understood well, it is mentioned [here](https://github.com/RTKConsortium/RTK/blob/master/include/rtkDualEnergyNegativeLogLikelihood.h#L112) that the variance of the integrated energy is assumed to be equal to the mean. Then, the function [GetVariances](https://github.com/RTKConsortium/RTK/blob/master/include/rtkDualEnergyNegativeLogLikelihood.h#L125-L139) is computing this:
``` math
\int_0^\infty E S_{eff}(E)exp(-\sum_{\alpha=1}^{M=2} f_\alpha A_\alpha)dE
```
which correspond to the mean signal measured by an energy integrating detector. 
(This is my understanding but I'm not sure I'm right.)

arobert01:
The last force-pushed improve the way the computation of the Cramer-Rao lower bound was implemented and add the computation of the variance also for spectral CT.

PR #621, opened 2024-09-24
ENH: Reorder the photon counts inside one-step filter

arobert01:
When subsets are used, the photon counts are shuffled before doing the reconstruction. The shuffle was done in the application and not in the filter which is not how it is done for the others reconstruction filters.

### DIFF
--- a/include/rtkMechlemOneStepSpectralReconstructionFilter.hxx
+++ b/include/rtkMechlemOneStepSpectralReconstructionFilter.hxx
@@ -57,6 +57,8 @@ MechlemOneStepSpectralReconstructionFilter<TOutputImage, TPhotonCounts, TSpectru
   m_NewtonFilter = NewtonFilterType::New();
   m_NesterovFilter = NesterovFilterType::New();
   m_MultiplySupportFilter = MultiplyFilterType::New();
+  m_ReorderPhotonCountsFilter = ReorderProjectionsFilterPhotonCountsType::New();
+  m_ReorderProjectionsWeightsFilter = ReorderProjectionsFilterProjectionsWeightsType::New();
 
   // Set permanent parameters
   m_ProjectionsSource->SetConstant(itk::NumericTraits<typename TOutputImage::PixelType>::ZeroValue());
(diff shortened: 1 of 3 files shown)


include/rtkMechlemOneStepSpectralReconstructionFilter.hxx
```diff
@@ -427,6 +449,17 @@ MechlemOneStepSpectralReconstructionFilter<TOutputImage, TPhotonCounts, TSpectru
...
+  // If subsets are used the photons counts and geometry are shuffled, then geometry need to be set again.
+  if (m_NumberOfSubsets != 1)
+  {
+    m_ReorderPhotonCountsFilter->Update();
+    this->SetGeometry(m_ReorderPhotonCountsFilter->GetOutputGeometry());
+    m_ForwardProjectionFilter->SetGeometry(this->m_Geometry);
+    m_SingleComponentForwardProjectionFilter->SetGeometry(this->m_Geometry);
+    m_GradientsBackProjectionFilter->SetGeometry(this->m_Geometry.GetPointer());
+    m_HessiansBackProjectionFilter->SetGeometry(this->m_Geometry.GetPointer());
+  }
```
SimonRit: When I look at the code of `ReorderProjectionsImageFilter`, I don't see why this is required. It should be sufficient to do it in UpdateOutputInformation no?

SimonRit (commented):
There is a bug, see [here](https://github.com/RTKConsortium/RTK/actions/runs/11015583587/job/30606183293#step:12:183). Do you know what could be the issue? And could you modify the graph to include the new filters in the class doc? Thanks!

include/rtkMechlemOneStepSpectralReconstructionFilter.hxx
arobert01: I tried by setting the new geometry only in the `GenerateInputRequestedRegion()` method but then I get this error `Mismatch between the number of projections and the geometry entries. Geometry has 0 entries, which is less than the last index of the projections stack`. Maybe I'm missing something here.

arobert01:
The last force-pushed fix the bug and update the documentation.

include/rtkMechlemOneStepSpectralReconstructionFilter.hxx
SimonRit: Reading [section 8.3.2 of the software guide](https://itk.org/ITKSoftwareGuide/html/Book1/ITKSoftwareGuide-Book1ch8.html#x54-1320008.3), I think the checks on the geometry should be in the `VerifyInputInformation` and not in the `VerifyPreconditions` so that the geometry can be modified in the `UpdateOutputInformation`. Does it tell you which filter complains with this error message?

arobert01: Yes the error comes from the `BackProjectionImageFilter` [here](https://github.com/RTKConsortium/RTK/blob/master/include/rtkBackProjectionImageFilter.hxx#L56).

SimonRit: If we want to avoid doing this, we need to rewrite `ReorderProjectionsImageFilter` to create the geometry in `GenerateOutputInformation`. That's for a different PR I guess, thanks for finding the proposed solution.

SimonRit (approved):

PR #622, opened 2024-09-26
ENH: Update geometry and signal in GenerateOutputInformation

SimonRit:
A filter generating a geometry should do it in GenerateOutputInformation

### DIFF
--- a/include/rtkMechlemOneStepSpectralReconstructionFilter.hxx
+++ b/include/rtkMechlemOneStepSpectralReconstructionFilter.hxx
@@ -341,6 +341,10 @@ MechlemOneStepSpectralReconstructionFilter<TOutputImage, TPhotonCounts, TSpectru
     m_ReorderPhotonCountsFilter->SetInputGeometry(this->m_Geometry);
     m_ReorderPhotonCountsFilter->SetPermutation(ReorderProjectionsFilterPhotonCountsType::SHUFFLE);
     m_ExtractPhotonCountsFilter->SetInput(m_ReorderPhotonCountsFilter->GetOutput());
+    m_ForwardProjectionFilter->SetGeometry(m_ReorderPhotonCountsFilter->GetOutputGeometry());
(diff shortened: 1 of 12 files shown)


include/rtkMechlemOneStepSpectralReconstructionFilter.hxx
```diff
@@ -341,6 +341,10 @@ MechlemOneStepSpectralReconstructionFilter<TOutputImage, TPhotonCounts, TSpectru
     m_ReorderPhotonCountsFilter->SetInputGeometry(this->m_Geometry);
     m_ReorderPhotonCountsFilter->SetPermutation(ReorderProjectionsFilterPhotonCountsType::SHUFFLE);
     m_ExtractPhotonCountsFilter->SetInput(m_ReorderPhotonCountsFilter->GetOutput());
+    m_ForwardProjectionFilter->SetGeometry(m_ReorderPhotonCountsFilter->GetOutputGeometry());
+    m_SingleComponentForwardProjectionFilter->SetGeometry(m_ReorderPhotonCountsFilter->GetOutputGeometry());
+    m_GradientsBackProjectionFilter->SetGeometry(m_ReorderPhotonCountsFilter->GetOutputGeometry());
+    m_HessiansBackProjectionFilter->SetGeometry(m_ReorderPhotonCountsFilter->GetOutputGeometry());
```
arobert01, line 348: If we put this here the change of geometry will be overriden few lines after ([here](https://github.com/RTKConsortium/RTK/blob/master/include/rtkMechlemOneStepSpectralReconstructionFilter.hxx#L430-L433)).

SimonRit: It's a good point and it's actually surprising that this has not been caught up by the tests... It's been fixed by moving the overriding code to the else.

arobert01:
Looks good to me

PR #632, opened 2024-11-08
ENH: Add variance reconstruction for FDK

arobert01:


SimonRit (changes requested):
Thanks for taking care of this overdue PR. A few comments:
- The command line application is wrong for short scans and displaced detector as you do not square the weights. I would suggest to at least remove them. Are you sure the command line application should be kept? Maybe a Python command line application is sufficient?
- There is also a test failing: https://my.cdash.org/viewTest.php?onlyfailed&buildid=2717834
- I would try make the test even faster, maybe 10 projections only? It does not really matter, does it?

arobert01:
Thanks for your comments. 

In the last force-pushed I deleted the application, fixed the failed test and modified the test to be faster.

arobert01:
With the last force-pushed I implemented the idea in the first part of this [comment](https://github.com/RTKConsortium/RTK/pull/404#discussion_r601238011) to avoid code repetition.

PR #634, opened 2024-11-15
 Standardize Phantom Format to Forbild

axel-grc:
Summary: This merge request updates RTK to use only the Forbild phantom format for geometry definitions. All other phantom formats have been deprecated or removed for consistency and simplicity.

Key Changes:
-Remove option SetIsForbildConfigFile in rtkprojectgeometricphantom and rtkdrawgeometricphantom
-Create documentation on how to create a Phantom in the forbild format.
-Adapt FDK example with new forbild phantom format


acoussat:
It looks like your pull request is malformed, as it contains commits that were already submitted in other pull requests (for instance e401be8 which comes from #633). I think you should rebase your branch `forbild-as-default` back onto `master` while dropping the unnecessary commits on the way, just to keep the pull requests atomic. Feel free to ask if you need help in doing so!

PR #669, opened 2025-01-10
ENH: Add new Python example for inline/on-the-fly FDK reconstruction

SimonRit:


LANGZHIZHEN:
_reader. UpdateOutputInformation()_ to produce **ITK ERROR: MetaImageIO x5d2bc29ff920 (0) : the File always be read: ./output/projection_049. mha for reading**. Is it because the reader reads the names of all the projected images?

SimonRit:
> _reader. UpdateOutputInformation()_ to produce **ITK ERROR: MetaImageIO x5d2bc29ff920 (0) : the File always be read: ./output/projection_049. mha for reading**. Is it because the reader reads the names of all the projected images?

Sorry, I'm not sure what it worked at some point but the ProjectionsReader uses the ImageSeriesReader which itself uses the first and the last projection. The last force-push fixed this.

LANGZHIZHEN:
Thanks, I can use it to develop my program. Best Regards!

LANGZHIZHEN:
> > _reader. UpdateOutputInformation()_ to produce **ITK ERROR: MetaImageIO x5d2bc29ff920 (0) : the File always be read: ./output/projection_049. mha for reading**. Is it because the reader reads the names of all the projected images?
> 
> Sorry, I'm not sure what it worked at some point but the ProjectionsReader uses the ImageSeriesReader which itself uses the first and the last projection. The last force-push fixed this.

When I was developing the cuda version, I had a problem. Is it possible for the above code to implement CUDA version? 
ITK ERROR: ExtractImageFilter(0x574e9f1f45f0): The number of zero sized dimensions in the input image Extraction Region
is not consistent with the dimensionality of the output image.
Expected the extraction region size ([0, 0, 0]) to contain 0 zero sized dimensions to collapse.

SimonRit:
Answered on https://discourse.itk.org/t/projection-images-config/7380/5. Please stop duplicating your posts.

SimonRit:
This code has been integrated into the repository in #680.

PR #679, opened 2025-02-05
Fix compilation of remote modules that depend on RTK in github python builds workflow

LucasGandel:
- Generated headers should be configured in the include subfolder
- The CMAKE_ARCHIVE_OUTPUT_DIRECTORY should default to the lib folder of the ITK build tree
- RTK_BUILD_APPLICATIONS should be exported for other modules that need to know its value at configure time

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -136,7 +136,14 @@ if(NOT CMAKE_LIBRARY_OUTPUT_DIRECTORY)
   set(CMAKE_LIBRARY_OUTPUT_DIRECTORY ${RTK_BINARY_DIR}/${CMAKE_INSTALL_LIBDIR})
 endif()
 if(NOT CMAKE_ARCHIVE_OUTPUT_DIRECTORY)
-  set(CMAKE_ARCHIVE_OUTPUT_DIRECTORY ${RTK_BINARY_DIR}/${CMAKE_INSTALL_LIBDIR})
+  # Python builds CI workflows expects archives to be placed in the ITK lib
+  # directory. This makes sure that other remote modules that depends on RTK
+  # can link against the RTK libraries.
+  if (WRAP_ITK_INSTALL_COMPONENT_IDENTIFIER STREQUAL "PythonWheel")
+    set(CMAKE_ARCHIVE_OUTPUT_DIRECTORY ${ITK_DIR}/${CMAKE_INSTALL_LIBDIR})
+  else()
+    set(CMAKE_ARCHIVE_OUTPUT_DIRECTORY ${RTK_BINARY_DIR}/${CMAKE_INSTALL_LIBDIR})
+  endif()
 endif()
 
 #=========================================================
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -173,6 +173,9 @@ list(APPEND RTK_INCLUDE_DIRS
 # Generate RTKConfig.cmake for the build tree.
 set(RTK_MODULE_PATH_CONFIG ${CMAKE_MODULE_PATH})
 
+# Add option to control the generation of RTK applications
+option(RTK_BUILD_APPLICATIONS "Build RTK applications" ON)
+
 set(RTK_EXPORT_CODE_BUILD "
 # The RTK version number
 set(RTK_VERSION_MAJOR ${RTK_VERSION_MAJOR})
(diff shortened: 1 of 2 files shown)


LucasGandel:
@SimonRit this change is required to successfully compile python packages of remote modules that depend on RTK (such as PCT)

CMakeLists.txt
```diff
@@ -136,7 +136,14 @@ if(NOT CMAKE_LIBRARY_OUTPUT_DIRECTORY)
...
 endif()
 if(NOT CMAKE_ARCHIVE_OUTPUT_DIRECTORY)
-  set(CMAKE_ARCHIVE_OUTPUT_DIRECTORY ${RTK_BINARY_DIR}/${CMAKE_INSTALL_LIBDIR})
+  # Python builds CI workflows expects archives to be placed in the ITK lib
+  # directory. This makes sure that other remote modules that depends on RTK
+  # can link against the RTK libraries.
+  if (WRAP_ITK_INSTALL_COMPONENT_IDENTIFIER STREQUAL "PythonWheel")
+    set(CMAKE_ARCHIVE_OUTPUT_DIRECTORY ${ITK_DIR}/${CMAKE_INSTALL_LIBDIR})
+  else()
+    set(CMAKE_ARCHIVE_OUTPUT_DIRECTORY ${RTK_BINARY_DIR}/${CMAKE_INSTALL_LIBDIR})
```
SimonRit, line 145: Is it required to keep the original directory if we're not building Python wrappings?

CMakeLists.txt
```diff
@@ -190,6 +203,9 @@ set(RTK_VERSION_PATCH ${RTK_VERSION_PATCH})
 # Whether the compiled version of RTK uses CUDA
 set(RTK_USE_CUDA ${RTK_USE_CUDA})
 
+# Whether RTK applications were built
+set(RTK_BUILD_APPLICATIONS ${RTK_BUILD_APPLICATIONS})
+
```
SimonRit, line 208: I'm not following why this is set three times (lines 177, 189 and 207). Isn't line 177 sufficient? If not, an explanatory comment would be appreciable for future reviews of the file...

SimonRit (approved):
Thanks, it looks like it addresses painful issues. I just have a few inline questions but I'm happy to merge it as is if all tests pass.

CMakeLists.txt
LucasGandel: The original directory can be used when building Python wrappings locally, it is just required to change it when building them with the ITKPythonPackage scripts to make sure the library is built in a cached directory. WRAP_ITK_INSTALL_COMPONENT_IDENTIFIER is only set to "PythonWheel" in the cmake options of the ITKPythonPackage scripts, so it should be safe.

l. 131 to l.146 set the default directories to the RTK build tree. By default other remote modules do not define anything and libraries are placed in the ITK build tree thanks to [this logic](https://github.com/InsightSoftwareConsortium/ITK/blob/master/CMake/ITKModuleExternal.cmake#L90). Because ITKModuleExternal is included after generating lpsolve, we must define the default directory values anyway, to make sure everything can be found in the same directory. I like the fact that RTK puts its libs in its build directory, this is what I would expect from a module built externally.

CMakeLists.txt
LucasGandel: Line 177 is enough to define the CMake option when building RTK. 189 and 207 are defined in the export strings (viewing this on github diff does not help to clearly identify that it is part of a string and not in the RTK CMakeLists code). The export strings are used to pass information to other modules, the same way it's done for RTK_VERSION_* and RTK_USE_CUDA above.

SimonRit: My bad, I reviewed this too quickly. Thanks for the explanation and sorry for the noise.

LucasGandel:
> Thanks, it looks like it addresses painful issues.

Indeed 😃 The problem was mainly caused by the ITKPythonPackage scripts on Linux which cache only a part of the dependency module, and RTK is now a dependency of PCT and has a complex build system (lpsolve third party, configured headers,...)

Modifications to the IPP scripts are also required to build on macOS, I'll open an issue on the PCT repo to describe what is still to be done

SimonRit:
Can you rebase on master? I have updated the CI in #678, it should pass after rebasing.

LucasGandel:
> Can you rebase on master? I have updated the CI in #678, it should pass after rebasing.

Outstanding work fixing the clang-format action 💯 🥇 🚀 

Rebased on master.

PR #680, opened 2025-02-10
Documentation improvement

axel-grc:
Modifications:

- add new code example inline reconstruction
- add tabs to separate python and c++ codes to more readability
- move the images to the documentation folder 

To do:

[ ] Add code examples from the forums
[ ] Create Python application of FDK


SimonRit:
@axel-grc I have force-pushed a few changes, notably rtkfdk.py which is not ready to be merged.

PR #683, opened 2025-02-12
Improve interoperability of spectral filters

cyrilmory:
Changed the image type of the incident spectrum in two-steps spectral filters to match that of the one-step filter (backward-compatible thanks to overloaded SetInput functions)
Overloaded the SetInput functions of the spectral one-step filter to accept the image type used by the two-steps filters.


SimonRit:
@cyrilmory In addition to rebasing on master, I have rewritten the history in two commits. Diff with your old commit is here:
https://github.com/RTKConsortium/RTK/compare/447fee1ac7d9c47e066db1ce7ccd2e68cfa8edce..0681326536bcf51681822a207b22d6f57e8598f6
Now, in the review process, I realized that "incident" is the French word for "impinging"... Can we rename all incident with impinging? You can amend the last commit for this I guess.
Thanks!
Simon

SimonRit:
"Incident spectrum" is actually widely used too so we keep it as is.

PR #684, opened 2025-02-18
Fix wrapping of double & output arguments

acoussat:
The problem is that in the method `rtkConvexShape::IsIntersectedByRay(const PointType &, const VectorType &, double &, double &)`, the last two parameters are meant to be output values, and thus cannot be used in Python, since there is no way to define a `double &` in Python.

The solution, crafted thanks to the pointers given by @SimonRit and [this comment](https://discourse.itk.org/t/additional-python-types-for-itk-remote-external-modules/2509/8) from @LucasGandel, removes the last two parameters from the method in Python and replaces them with output values (see example below). There might be a better way of achieving the same result, but I am far from fluent in SWIG, so any suggestion of improvement is welcome!

Test script:
```python
import itk
from itk import RTK as rtk

q = rtk.QuadricShape.New()

# Define a sphere of radius 10
radius = 10
q.SetA(1)
q.SetB(1)
q.SetC(1)
q.SetJ(-radius**2)

print(q.IsIntersectedByRay([-20,0,0],[1,0,0]))  # only two parameters
```

Output with the fix:
```
[True, 10.0, 30.0]
```

### DIFF
--- a/test/CMakeLists.txt
+++ b/test/CMakeLists.txt
@@ -306,4 +306,5 @@ rtk_add_test(rtkBioscanTest rtkbioscantest.cxx
 if(ITK_WRAP_PYTHON)
   itk_python_add_test(NAME rtkFirstReconstructionPythonTest COMMAND rtkFirstReconstruction.py ${CMAKE_CURRENT_BINARY_DIR}/rtkFirstReconstruction.mha)
   itk_python_add_test(NAME rtkMaximumIntensityPythonTest COMMAND rtkMaximumIntensity.py  ${CMAKE_CURRENT_BINARY_DIR}/rtkFirstReconstruction.mha)
+  itk_python_add_test(NAME rtkOutoutArgumentWrappingTest COMMAND rtkOutoutArgumentWrapping.py)
 endif()
--- a/test/rtkOutputArgumentWrapping.py
+++ b/test/rtkOutputArgumentWrapping.py
@@ -1,0 +1,44 @@
+import itk
+from itk import RTK as rtk
+
+# Test for rtkConvexShape::IsIntersectedByRay
+
+q = rtk.QuadricShape.New()
+
+# Define a sphere of radius 10
+radius = 10
+q.SetA(1)
+q.SetB(1)
+q.SetC(1)
+q.SetJ(-radius**2)
+
+print(q.IsIntersectedByRay([-20, 0, 0], [1, 0, 0]))
+
+# Test for rtk::FieldOfViewImageFilter::ComputeFOVRadius
+
+geometry = rtk.ThreeDCircularProjectionGeometry.New()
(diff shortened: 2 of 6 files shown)


SimonRit (changes requested):
Thanks, great work. I'm not fluent in swig so I trust it's very well written :-). However, I speak a bit of cmake and it is meaning less to use `configure_file` here as there is nothing to configure (cmake variables with @, see other `.in` files)... Can you rename `ConvexShape.i.in` to `ConvexShape.i` and replace `${CMAKE_CURRENT_BINARY_DIR}/ConvexShape.i` by `${CMAKE_CURRENT_SOURCE_DIR}/ConvexShape.i` please?

acoussat:
Thanks for the comment. As we discussed, we can also implement the same fix for all the relevant `double &` in RTK. You told me about the ones in `rtkFieldOfViewImageFilter::ComputeFOVRadius`, shall we prepare a small list? We can also deal with `int &` or `float &` if they also appear.

SimonRit:
> Thanks for the comment. As we discussed, we can also implement the same fix for all the relevant `double &` in RTK. You told me about the ones in `rtkFieldOfViewImageFilter::ComputeFOVRadius`, shall we prepare a small list? We can also deal with `int &` or `float &` if they also appear.

I used the following search `grep -rE '\s*(short|double|int|float|string) *&' include/*h -B2` and found in addtion to IsIntersectedByRay (excluding const ones)
- rtk::FieldOfViewImageFilter::ComputeFOVRadius
- rtk::ThreeDCircularProjectionGeometry::FixAngles

SimonRit:
Please also rebase on master to fix the CI.

acoussat:
I haven't completely tested it, maybe I first try to make sure that it works properly before merging it.

acoussat:
After further testing, `rtk::FieldOfViewImageFilter::ComputeFOVRadius` seems to properly work. However, `rtk::ThreeDCircularProjectionGeometry::FixAngles` does not (the method does not seem to be wrapped). After investigation, I think this is because `FixAngles` is declared as `protected`, so it is not meant to be used outside of the class itself.
The solution is either to:
* not wrap `FixAngles` (I think this is better)
* make `FixAngles` `public`

Let me know what you think.

Also, I have a small script that tests the features of the methods detailed above. Let me know if you are interested in it and what to do with it. I think that once Python automatic testing is set-up, it could be nice to run it in order to avoid future regressions.

acoussat:
I added a small Python test and squashed all commits into [c580868](https://github.com/RTKConsortium/RTK/pull/684/commits/c580868df89a199bc6ce551686668edd234d50dd).

SimonRit:
Thanks! The test is currently failing but this is an independent problem I believe

PR #691, opened 2025-02-21
DOC: Only recompile Sphinx documentation when required

SimonRit:


SimonRit:
What do you think @LucasGandel?

SimonRit:
My initial attempt was polluting the source directory. The latest force push doesn't but it also doesn't clean up the compiled doc anymore. I'm open to any suggestion!

LucasGandel:
Not sure if there is a clean way to do that. I'll discuss with CMake experts and will update here

LucasGandel:
According to other CMake experts, it looks like the only way is to check each files in the directory...

SimonRit:
@LucasGandel I have force pushed a new suggestion. This is not a simple piece of code but I don't know a way of doing it better. Do you think it would do?

LucasGandel:
> @LucasGandel I have force pushed a new suggestion. This is not a simple piece of code but I don't know a way of doing it better. Do you think it would do?

Thanks for doing this. From a quick glance it looks good to me, everything is done in a single file which will make it easier to maintain.

PR #709, opened 2025-04-01
BUG: Fix typo in OutputArgumentWrappingTest

acoussat:
PR #684 introduced a silly typo.


SimonRit:
Ok, we need to enable Python tests on the CI...

acoussat:
I agree, is it complicated to do? Is it enough to enable `BUILD_TESTING` there: https://github.com/RTKConsortium/RTK/blob/a5ead44eacc05352fc8d95c0c1fd6837c32b08e0/.github/workflows/build-test-package.yml#L16?

SimonRit:
No it's not enough, you need to enable the wrappings as well in ITK. It would be better to reuse the existing wrapping CI. It's not straightforward but it might be not too difficult. See the discussion in RTKConsortium/ITKCudaCommon#53.

PR #731, opened 2025-05-07
ENH: Create python API for applications

axel-grc:
**Enable use of applications via the Python API**
Modified rtkfdk to support both CLI usage and direct calls from Python. Now supports:

- rtk.rtkfdk("-p . -r projections.mha -o fdk.mha -g geometry.xml --spacing 2 --dimension 256")
- rtk.rtkfdk(path=".", regexp="projections.mha", output="fdk.mha", geometry="geometry.xml", spacing=2, dimension=256, verbose=True)

Signature is auto-patched for better help and to raise standard Python exceptions instead of argparse errors.

### DIFF
--- a/applications/README.md
+++ b/applications/README.md
@@ -1,6 +1,8 @@
-# Command-line examples
+# Command-line applications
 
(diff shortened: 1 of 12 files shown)


axel-grc:
Can you validate @SimonRit ?
If it's ok I can do the same for all applications

SimonRit:
To be changed:
- Make pyvista dependency optional
- Display (shell-style) help of each application
- Merge new page doc to "Command-line examples" and rename it to "Command-line applications"

SimonRit (approved):

PR #754, opened 2025-06-13
ENH: Add rtkadmmtotalvariation python application

axel-grc:
WIP: More test needed after changing the wrap file

### DIFF
--- a/applications/rtkadmmtotalvariation/rtkadmmtotalvariation.py
+++ b/applications/rtkadmmtotalvariation/rtkadmmtotalvariation.py
@@ -1,0 +1,185 @@
+#!/usr/bin/env python
+import argparse
+import sys
+import itk
+from itk import RTK as rtk
+
+
+def build_parser():
+    # Argument parsing
+    parser = argparse.ArgumentParser(
+        description="Perform ADMM total variation reconstruction on cone-beam projections.",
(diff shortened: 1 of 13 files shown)


SimonRit:
What's the status of this PR? Have you fixed the issue we discussed at the RTK training?

axel-grc:
Not yet, I'll work on that after the PR I'm working on.

axel-grc:
When testing with cuda, I encounter a problem when running the command line example (it works with cpu) :
```
rtkadmmtotalvariation -p . -r projections.mha -o admmtv.mha -g geometry.xml --spacing 2 --size 256 --alpha 1 --beta 1000 -n 3
```
I get this error :
```
RuntimeError: /work/ITK-source/ITK/Modules/Core/Common/include/itkImageConstIterator.h:210:
ITK ERROR: Region ImageRegion (0x7ffe561b5270)
  Dimension: 3
  Index: [0, 224, 179]
  Size: [256, 32, 1]
 is outside of buffered region ImageRegion (0x28f21f28)
  Dimension: 3
  Index: [0, 0, 0]
  Size: [0, 0, 0]
```

The problem does not seem to come from the inputs.
Any ideas @SimonRit ?

SimonRit:
Problem solved with the last force push, which is only a patch but having the same behavior with Python would require solving RTKConsortium/ITKCudaCommon#52.

Issue #639, opened 2024-12-04
Command line applications should be translated to Python


SimonRit:
This is the current list of command line applications (crosse-out ones have been / should be removed):
- [x] rtkadmmtotalvariation
- [x] rtkadmmwavelets
- [x] rtkamsterdamshroud
- [x] rtkbackprojections
- [x] rtkbioscangeometry
- [x] rtkcheckimagequality
- [x] rtkconjugategradient
- [x] rtkdigisensgeometry
- [x] rtkdrawgeometricphantom
- [x] rtkdrawshepploganphantom
- [x] rtkdualenergyforwardmodel
- [x] rtkdualenergysimplexdecomposition
- [x] rtkelektasynergygeometry
- [x] rtkextractphasesignal
- [x] rtkextractshroudsignal
- [x] rtkfdk
- [ ] ~~rtkfdktwodweights~~
- [x] rtkfieldofview
- [x] rtkforwardprojections
- [x] rtkfourdconjugategradient
- [x] rtkfourdfdk
- [x] rtkfourdrooster
- [x] rtkfourdsart
- [x] rtkgaincorrection
- [x] rtki0estimation
- [x] rtkimagxgeometry
- [x] rtkiterativefdk
- [x] rtklagcorrection
- [ ] ~~rtklastdimensionl0gradientdenoising~~
- [ ] ~~rtklut~~
- [x] rtkmaskcollimation
- [ ] ~~rtkmcrooster~~
- [ ] ~~rtkmotioncompensatedfourdconjugategradient~~
- [x] rtkorageometry
- [x] rtkosem
- [x] rtkoverlayphaseandshroud
- [ ] ~~rtkparkershortscanweighting~~
- [x] rtkprojectgeometricphantom
- [ ] rtkprojectionmatrix
- [x] rtkprojections
- [x] rtkprojectshepploganphantom
- [ ] ~~rtkrabbitct~~
- [ ] ~~rtkramp~~
- [ ] ~~rtkrayboxintersection~~
- [ ] ~~rtkrayellipsoidintersection~~
- [ ] ~~rtkrayquadricintersection~~
- [x] rtkregularizedconjugategradient
- [x] rtksart
- [x] rtkscatterglarecorrection
- [x] rtksimulatedgeometry
- [x] rtkspectraldenoiseprojections
- [x] rtkspectralforwardmodel
- [x] rtkspectralonestep
- [x] rtkspectralrooster
- [x] rtkspectralsimplexdecomposition
[...]

PR #758, opened 2025-06-19
WIP: Display version in Python command line tools

### LINKED ISSUE #722: Python command line tools should display the version, as C++ command line tools
See https://github.com/RTKConsortium/RTK/blob/master/applications/rtkVersion.ggo.in which is processed [here](https://github.com/RTKConsortium/RTK/blob/master/applications/CMakeLists.txt#L12) and included in, e.g., [CMakeLists.txt](https://github.com/RTKConsortium/RTK/blob/master/applications/rtkfdk/CMakeLists.txt)

acoussat:
This is what I came up with to fix #722 during the hackathon. I only converted one example, namely `rtkdrawgeometricphantom`, but if the logic is satisfactory then I will convert the other Python applications (hence the WIP).

Advantages of this approach:
- No additional source of truth for the version, so it should always be consistent with the C++ code
- Single source of truth for the definition of the `ArgumentParser`, meaning that if one day we want to change the format of the `--help` command, then we have only one file to edit
- Nice side effect, there is now a new function to check RTK's version directly from Python using `rtk.version()`

Let me know what you think.


axel-grc:
Thanks for your contribution.
I don't think it is necessary to create a rtkversion.py file or modify the CMakeLists of the application folder.
What I suggest to do is to create a class inherited from argparse.ArgumentParser and in the init add the argument of the version.

For example:

``` python
import argparse
from importlib.metadata import version
class argument_parser(argparse.ArgumentParser):
    def __init__(self, *args, version=None, **kwargs):
        super().__init__(*args, **kwargs)
        self..add_argument('--version', action='version', version=version("itk-rtk"))
```

With a custom parser,  we could move some options from rtkExtras file such as comma_separated_args directly into this new parser.

acoussat:
Thanks for your proposal! I have given some thoughts about your idea of subclassing `ArgumentParser` but I am not sure if I see how to implement `comma_separated_args` directly in the parser. I have also tried to use `version("itk-rtk")` as in your example but it does not work with self-compiled RTK, and it does not print the commit hash as it does with the ggo version. Perhaps we can have a small discussion about that later on. Anyway, feel free to suggest changes directly in this PR, you should have the rights to do so!

SimonRit:
I think you need to add `configure_file` in `wrapping/CMakeLists.txt`. I would keep `rtkversion.py.in` in  `${RTK_SOURCE_DIR}/applications` and `rtkversion.py` in `${RTK_BINARY_DIR}/applications`.

acoussat:
Thanks, it seems to work now! If the code looks good to you, I would like to convert all remaining Python applications to include the version number. Let me know when I can proceed.

SimonRit:
I have been thinking about it and in our endeavor to match the C++ version and the Python version, I'm not too fond of the use of epilog. Instead, I would suggest to use `usage` and add a `--version` / `-V` argument. Were there any counter arguments against this?
BTW, the git hash didn't display when I installed the wheel.

axel-grc:
Just for your information, I found the library [setuptools-scm](https://pypi.org/project/setuptools-scm/), which can generate a _version.py file. However, this would add another dependency, which I don't think is necessary in our case.

I don't have a strong opinion about whether the version should be displayed in the epilog or the usage message.

In the future, I would like to create a new parser class that inherits from ArgumentParser, rather than just using a function.

Would you like me to make a new proposal?

SimonRit:
Discuss with @acoussat whether it's better to merge this temporary solution or propose a new one.

acoussat:
I guess your new proposal would supersede what I did in this PR, so I think you can create a new PR with your proposal (starting either from scratch or from this PR) then we can compare and close one of the two. Thanks!

SimonRit:
Let's finish this PR first:
- Define the git hash in the root `CMakeLists.txt`
- Create an option `--version`, `-V` of type `version` for each application
- Try to modify the formatter to have the version at the top, then description and finally usage

acoussat:
Thanks, I've implemented the changes into separate commits to make reviewing easier. I'll squash the commits once you approve the code.

Two additional comments:
- I order to implement commit e9c9c9b I followed the trick explained [here](https://stackoverflow.com/a/22382835). It's not the prettiest, but it implements a new formatter, as was suggested by @axel-grc 
- While implementing commit 4a142f2 I have noticed that many Python applications have unused `import` statements, maybe a clean-up would be needed, should I open an issue (or do it myself directly)?

axel-grc:
Looks good to me @acoussat 
I copied the imports across all python application, so yes there is some that are not used, you can remove them or I could do it in another PR

acoussat:
I would clean the `import` statements once #639 is completed. In the meantime I can create an issue so that we remember. What do you think?

axel-grc:
Yes you can create an issue. 
Thanks

SimonRit:
Addressed by  #792

PR #771, opened 2025-07-02
STYLE: remove “using” aliases to inline single-use types

axel-grc:
My script is maybe a little too aggressive, for example, it removes "using OutputPixelType = double" .
But no compilation error and all test passed.


axel-grc:
The script is not perfect and there where some errors that I fixed manually.

Ready to merge

Issue #773, opened 2025-07-03
Remove unnecessary templates


SimonRit:
Several classes are using template parameters which don't seem useful to me, for example [rtk::ADMMTotalVariationConeBeamReconstructionFilter ](https://github.com/RTKConsortium/RTK/blob/master/include/rtkADMMTotalVariationConeBeamReconstructionFilter.h#L134) defines a TGradientOutputImage. When the default value is always used and will likely always be used, we should remove these template classes before releasing RTK v3.

axel-grc:
close #782

PR #782, opened 2025-07-21
STYLE: Remove unused template parameters

axel-grc:


SimonRit:
Nice work, thanks! As a followup, it would be nice to have all iterative filters provide the same templates as the mother class for consistency:
```c++
 template <class TOutputImage, class ProjectionStackType = TOutputImage>
```

axel-grc:
Closes #773

PR #788, opened 2025-07-29
Python test

### LINKED ISSUE #765: Testing should compile and run self-contained example projects
This would allow mistakes such as
https://github.com/RTKConsortium/RTK/compare/eea64c8895e8b38ec5771409043850b20b9e2474..ab089fc91b37dd19443499d85b0ae74922e6ce0b

axel-grc:
Closes #765


.github/workflows/build-test-package-python-cuda.yml
```diff
@@ -212,3 +212,33 @@ jobs:
...
+        python-version: '3.11'
+
+    - name: Test python wheel
+      run: |
+        # Find the CUDA 124 wheel for Python 3.11 dynamically
+        wheel=$(find wheels -name "*cuda124*cp311*manylinux_2_28_x86_64.whl" -type f | head -1)
+
+        pip uninstall -y $(pip list |  grep -E '^itk-')
+        echo "Installing wheel: $wheel"
+        pip install $wheel
```
acoussat, line 241: I know this is how ITK does it, but did you figure out why not directly use `pip install --force-reinstall`?

.github/workflows/build-test-package-python-cuda.yml
```diff
@@ -212,3 +212,33 @@ jobs:
...
+
+        pip install pytest
+        pytest $GITHUB_WORKSPACE/test/*.py -vv -W ignore
```
acoussat: I see that `-W ignore` is used to remove some cryptic Swig warnings, but it would be nice if we could filter out only those warnings and keep the other ones just in case. Then the line would become something like:
```bash
pytest $GITHUB_WORKSPACE/test/*.py -vv -W "ignore:builtin type swig"
```

examples/FirstReconstruction/FirstReconstruction.py
```diff
@@ -3,83 +3,86 @@
...
+    # Create a stack of empty projection images
+    ConstantImageSourceType = rtk.ConstantImageSource[ImageType]
+    constantImageSource = ConstantImageSourceType.New()
+    origin = [-127, -127, 0.0]
+    sizeOutput = [128, 128, numberOfProjections]
+    spacing = [2.0, 2.0, 2.0]
+    constantImageSource.SetOrigin(origin)
+    constantImageSource.SetSpacing(spacing)
+    constantImageSource.SetSize(sizeOutput)
+    constantImageSource.SetConstant(0.0)
```
acoussat: Out of curiosity, any specific reason why this bloc was moved?

test/rtkOutputArgumentWrapping.py
```diff
@@ -1,44 +1,44 @@
...
-constantImageSource.SetSize(sizeOutput)
-constantImageSource.SetConstant(0.0)
-source = constantImageSource.GetOutput()
-
-f = rtk.FieldOfViewImageFilter.New()
-f.SetGeometry(geometry)
-f.SetProjectionsStack(source)
-print(f.ComputeFOVRadius(f.FOVRadiusType_RADIUSINF))
+def test_OutputArgumentWrapping():
+    # Test for rtkConvexShape::IsIntersectedByRay
```
acoussat: Now that we run pytest, I would separate this test into two, one for `rtkConvexShape::IsIntersectedByRay`, and one for `rtk::FieldOfViewImageFilter::ComputeFOVRadius`.

test/rtkexamplestest.py
```diff
@@ -0,0 +1,42 @@
...
+        an.main()
+
+    def test_GeometricPhantomExample(tmp_path):
+        import GeometricPhantom.GeometricPhantom as gp
+        import urllib.request
+
+        url = "https://raw.githubusercontent.com/RTKConsortium/Forbild/refs/heads/main/Thorax"
+        thorax_path = tmp_path / "thorax.txt"
+        with urllib.request.urlopen(url) as response, open(thorax_path, "wb") as out_file:
+            out_file.write(response.read())
```
acoussat: Perhaps we could move these lines to a pytest fixture instead?

acoussat (commented):
Nice! Some minor comments but it looks like we're finally converging to a solution that works!

.github/workflows/build-test-package-python-cuda.yml
axel-grc: Yes, if we use --force-reinstall in the non-cuda wheel, it will keep the cuda packages

examples/AddNoise/AddNoise.py
```diff
@@ -72,4 +72,4 @@
 plt.ylabel("Pixel intensity")
 plt.legend()
 plt.savefig("AddNoise.png", bbox_inches="tight")
-plt
+#plt.show()
```
SimonRit: I would remove this line

examples/GeometricPhantom/GeometricPhantom.py
```diff
@@ -10,7 +11,7 @@
 sid = 600
 sdd = 1200
 scale = 2.0
-config_file_name = "Thorax"
+config_file_name = sys.argv[1]
```
SimonRit: Why change this? I would prefer to keep the example with an hardcoded filename.

test/rtktestexamples.py
```diff
@@ -0,0 +1,47 @@
...
+    script = EXAMPLES / rel_script
+    os.chdir(tmp_path)
+    sys.argv = [str(script), *map(str, args)]
+    runpy.run_path(str(script), run_name="__main__")
+
+
+@pytest.fixture
+def thorax_file(tmp_path):
+    url = "https://raw.githubusercontent.com/RTKConsortium/Forbild/refs/heads/main/Thorax"
+    thorax_path = tmp_path / "thorax.txt"
```
SimonRit: I guess you can just name it "Thorax" to accomodate the other comment on `GeometricPhantom.py`.

SimonRit (approved):
Very minor suggestions, it would be great to have @acoussat's approval too

test/CMakeLists.txt
```diff
@@ -306,7 +306,6 @@ rtk_add_test(rtkBioscanTest rtkbioscantest.cxx
 rtk_add_test(rtkVarianceReconstructionTest rtkvariancereconstructiontest.cxx)
 
 if(ITK_WRAP_PYTHON)
-  itk_python_add_test(NAME rtkFirstReconstructionPythonTest COMMAND rtkFirstReconstruction.py ${CMAKE_CURRENT_BINARY_DIR}/rtkFirstReconstruction.mha)
   itk_python_add_test(NAME rtkMaximumIntensityPythonTest COMMAND rtkMaximumIntensity.py  ${CMAKE_CURRENT_BINARY_DIR}/rtkFirstReconstruction.mha)
```
acoussat: Somewhat unrelated but I don't think `rtkMaximumIntensity.py` accepts any argument.

```suggestion
  itk_python_add_test(NAME rtkMaximumIntensityPythonTest COMMAND rtkMaximumIntensity.py )
```

test/rtkMaximumIntensity.py
```diff
@@ -2,39 +2,38 @@
...
-projInput.SetConstant(0.0)
-projInput.Update()
-projInputSource = projInput.GetOutput()
-# MIP Forward Projection filter
-mip = rtk.MaximumIntensityProjectionImageFilter[TImageType, TImageType].New()
-mip.SetGeometry(geometry)
-mip.SetInput(volInputSource)
-mip.SetInput(1, projInputSource)
-mipImage = mip.GetOutput()
+def test_MaximumIntensity(tmp_path):
```
acoussat: The fixture `tmp_path` is never used here.

test/rtkOutputArgumentWrapping.py
```diff
@@ -2,43 +2,45 @@
...
+    q.SetC(1)
+    q.SetJ(-(radius ** 2))
 
-# Define a sphere of radius 10
-radius = 10
-q.SetA(1)
-q.SetB(1)
-q.SetC(1)
-q.SetJ(-(radius ** 2))
+    print(q.IsIntersectedByRay([-20, 0, 0], [1, 0, 0]))
```
acoussat: Now that we have proper Python testing mechanisms, we should `assert` expected values instead of just printing whatever is returned by the function. This is perhaps a bit off-topic regarding the current PR, should I open another PR or do I let you @axel-grc deal with it?

acoussat (dismissed):
Nice to finally be able to run those Python tests! I just have some minor comments.

PR #799, opened 2025-08-27
ENH: Use RebindImageType to replace manual itk::Image vs itk::CudaImage selection

axel-grc:


include/rtkFourDROOSTERConeBeamReconstructionFilter.h
```diff
@@ -219,11 +219,37 @@ class ITK_TEMPLATE_EXPORT FourDROOSTERConeBeamReconstructionFilter
...
+                                                       CudaAverageOutOfROIImageFilter>;
+  using TemporalTVDenoisingFilterType =
+    std::conditional_t<std::is_same_v<VolumeSeriesType, CPUVolumeSeriesType>,
+                       TotalVariationDenoisingBPDQImageFilter<VolumeSeriesType, TemporalGradientImageType>,
+                       CudaLastDimensionTVDenoisingImageFilter>;
+#  else
+  using AverageOutOfROIFilterType = rtk::AverageOutOfROIImageFilter<VolumeSeriesType, VolumeType>;
+  using TemporalTVDenoisingFilterType =
+    rtk::TotalVariationDenoisingBPDQImageFilter<VolumeSeriesType, TemporalGradientImageType>;
+#  endif
```
SimonRit: Can't we avoid this `#ifdef RTK_USE_CUDA` using `Rebind`?

SimonRit (changes requested):
Looks good. However, those are style changes so replace ENH by STYLE in the commit messages. I also wonder if it's worth merging this or should we wait for a CudaCommon release?

include/rtkFourDROOSTERConeBeamReconstructionFilter.h
axel-grc: We can only use rebind on itk::Image or itk::CudaImage where the rebind alias is defined.
If we want to use rebind on AverageOutOfROIFilterType, I think we need to define rebind on AverageOutOfROIImageFilter and CudaAverageOutOfROIImageFilter

axel-grc:
I think it is better to wait for a CudaCommon release to merge

include/rtkFourDROOSTERConeBeamReconstructionFilter.h
SimonRit: I see, thanks. Does the code in `ifdef RTK_USE_CUDA` fail if `RTK_USE_CUDA` is `OFF`?

SimonRit (approved):

PR #800, opened 2025-08-27
DOC: Specified Tikhonov parameter in conjugate gradient reconstruction

IdrisCRS:


SimonRit:
Looks good, thanks. Can you just make the commit message shorter? See [here](https://docs.itk.org/en/latest/contributing/index.html): "Try to keep the subject line below 72 characters, ideally 50."
e.g. "DOC: Specified Tikhonov parameter in conjugate gradient reconstruction"

PR #802, opened 2025-08-28
Python app test

axel-grc:


SimonRit:
There seems to be a (real) CI issue with this one. Could you also rebase on master to avoid the style issue? Thanks!

SimonRit:
There is an issue visible on the CI:
https://github.com/RTKConsortium/RTK/actions/runs/17376156647/job/49335007969#step:5:128

wrapping/__init_rtk__.py
```diff
@@ -22,20 +21,27 @@
...
+    if not (fname.startswith("rtk") and fname.endswith(".py")):
+        continue
+    modname = fname[:-3]
+
+    if (modname.endswith("Python") or modname in _excluded_base_names):
+        continue
+    # Heuristic: application scripts are all lowercase (skip CamelCase bindings)
+    if any(c.isupper() for c in modname):
+        continue
+    app_names.append(modname)
```
SimonRit: This has become completely unreadable. What files were selected which shouldn't have? I suggest we roll back to the list instead of this.

SimonRit (approved):
I just force-pushed to fix a commit message (writter->writer), approved otherwise, I'll merge when the CI's green.

PR #804, opened 2025-08-29
BUG: Fix error no module named 'itk.rtkversion'

axel-grc:


SimonRit:
@axel-grc thanks. I just did a minor change which I force-pushed

SimonRit:
Thanks for this!

PR #810, opened 2025-09-09
Inline reconstruction

### LINKED ISSUE #809: InlineReconstruction diff in python and C++ (maybe a bug)
After I run the C++ example code, I found the result is not the same.
for first glance, I think C++ code has a bug.
In python , the result is 👍 and the result can make sence.

Image

but in C++, the result is not easy to understand, 3d slices are not like this, it doesn't make sence, neither the values.

Image

SimonRit:
Fixes #809


axel-grc:
Looks good @SimonRit , thanks.

PR #812, opened 2025-09-12
Warp projectors

axel-grc:
First draft to include warp projectors to iterative algos


SimonRit (changes requested):
Please remove the API change for ROOSTER

PR #813, opened 2025-09-16
ENH: Upgrade CI CUDA Python packages to CudaCommon v2

SimonRit:


SimonRit:
@LucasGandel We are in the process of testing RTKConsortium/ITKCudaCommon#43. I wanted to generate RTK packages compatible with this PR and it turns out that I need to modify ITKPythonPackages, see [here](https://github.com/SimonRit/ITKPythonPackage/commit/1d3c0287c2709327f40cc67136332f812b152d1b). I think we should move `cudaCommonConfiguration.h.in` to the include directory but for `CudaImage.i.in` and `CudaImage.i.init`, I don't have a better solution than patching ITKPythonPackage. What do you think?

SimonRit:
This is ready to be merged but the Python tests will fail until a new version of CudaCommon is released. I suggest we wait for the release of Python packages before merging it.

SimonRit:
We have compilation problems on main without this PR. Merging it, Python tests will fail until CudaCommon's release.

PR #816, opened 2025-09-22
BUG: Fix Python argument parser for Python 3.9

SimonRit:
Fixes error message
```
TypeError: unsupported operand type(s) for |: 'type' and 'NoneType'
```
This is because `str | None` syntax is only supported in Python 3.10 or later (https://peps.python.org/pep-0604/).

### DIFF
--- a/applications/rtkargumentparser.py
+++ b/applications/rtkargumentparser.py
@@ -3,6 +3,7 @@ import argparse
 from itk import RTK as rtk
 import difflib
 import inspect
+from typing import Optional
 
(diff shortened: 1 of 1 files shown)


acoussat:
Looks good to me, but makes me wonder why we have those type hints only in some parts of the code like here. Is there any reason to use them here? ITK does not seem to use them at all as far as I checked. @axel-grc maybe you would have some interesting input about this?

acoussat (approved):

PR #821, opened 2025-10-02
Improve ray quadric intersection calculations

### LINKED ISSUE #819: Mishandling of Forbild shapes in parallel geometry
Steps to reproduce:
Phantom file:
```
{
	[Cylinder_y: x=0 y=0 z=0 r=0.1  l=1 rho=1]
}
```
`rtksimulatedgeometry -n 180 -a 180 -o g.xml --sdd 0.`
`rtkprojectgeometricphantom -g g.xml -o p_acr.mha --phantomfile ACR_phantom.txt --size 201,201 --spacing 0.01`

And get:
```
ExceptionObject caught with ppc->Update() in file /Users/lesaint/src/RTK/applications/rtkprojectgeometricphantom/rtkprojectgeometricphantom.cxx line 103

itk::ExceptionObject (0x600003411c00)
Location: "unknown" 
File: /Users/lesaint/src/RTK/src/rtkQuadricShape.cxx
Line: 114
Description: Intersection of the quadric with the line gives two half lines, add clip planes to select which one
```

It works fine if:
- I use a divergent geometry.
- I change `spacing` or `size`.

It seems to systematically fail when the `size[0]`/`spacing` combination yields an origin at -1 sharp!

SimonRit:
Fixes #819.

### DIFF
--- a/src/rtkQuadricShape.cxx
+++ b/src/rtkQuadricShape.cxx
@@ -63,18 +63,52 @@ QuadricShape ::IsIntersectedByRay(const PointType &  rayOrigin,
   const ScalarType zero = itk::NumericTraits<ScalarType>::ZeroValue();
   if (Aq == zero)
   {
-    // Only one intersection with, e.g., a plane: check which half line should be kept
-    nearDist = -Cq / Bq;
-    // The following condition tests on which side of the quadric is the ray origin.
-    // It is equivalent to the following code.
-    // if ((IsInsideQuadric(rayOrigin) && nearDist < 0.) ||
-    //     (!IsInsideQuadric(rayOrigin) && nearDist > 0.))
-    if (Cq * nearDist > 0.)
-      farDist = itk::NumericTraits<ScalarType>::max();
+    if (Bq == zero)
+    {
+      if (Cq == zero)
+      {
+        // Whole ray contained in quadric
+        nearDist = -itk::NumericTraits<ScalarType>::max();
+        farDist = itk::NumericTraits<ScalarType>::max();
+      }
+      else
+        return false;
+    }
     else
     {
-      farDist = nearDist;
-      nearDist = -itk::NumericTraits<ScalarType>::max();
+      if (Cq == zero)
+      {
+        // rayOrigin is on the quadric, we need to find which half line is in
+        // the quadric For t=1, we have Aq+Bq+Cq, in this case Bq. As Aq and Cq
+        // are negative, we test the sign of Bq.
+        if (Bq > 0)
+        {
+          // rayOrigin+rayDirection is not in the quadric
+          farDist = 0.;
+          nearDist = -itk::NumericTraits<ScalarType>::max();
+        }
+        else
+        {
+          nearDist = 0.;
+          farDist = itk::NumericTraits<ScalarType>::max();
+        }
+      }
+      else
+      {
+        // Only one intersection with, e.g., a plane: check which half line should be kept
+        nearDist = -Cq / Bq;
+        // The following condition tests on which side of the quadric is the ray origin.
+        // It is equivalent to the following code.
+        // if ((IsInsideQuadric(rayOrigin) && nearDist < 0.) ||
+        //     (!IsInsideQuadric(rayOrigin) && nearDist > 0.))
+        if (Cq * nearDist > 0.)
+          farDist = itk::NumericTraits<ScalarType>::max();
+        else
+        {
+          farDist = nearDist;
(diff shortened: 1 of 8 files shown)


src/rtkQuadricShape.cxx
```diff
@@ -60,18 +60,35 @@ QuadricShape ::IsIntersectedByRay(const PointType &  rayOrigin,
...
+      nearDist = -Cq / Bq;
+      // The following condition tests on which side of the quadric is the ray origin.
+      // It is equivalent to the following code.
+      // if ((IsInsideQuadric(rayOrigin) && nearDist < 0.) ||
+      //     (!IsInsideQuadric(rayOrigin) && nearDist > 0.))
+      if (Cq * nearDist > zero)
+        farDist = itk::NumericTraits<ScalarType>::max();
+      else
+      {
+        farDist = nearDist;
```
lesaintjerome: Not sure to understand the meaning of nearDist and farDist : is it related to the distance to the source, or simply the smaller and the larger t_values ?

lesaintjerome: Shouldn't the case `Bq!=0 and Cq==0` be treated separately. In this case, one needs to search for the half-ray  (`t>0` or `t<0`) that is contained in the quadric.

src/rtkQuadricShape.cxx
```diff
@@ -84,7 +101,8 @@ QuadricShape ::IsIntersectedByRay(const PointType &  rayOrigin,
     {
       // No intersection but one might be dealing with an infinite line
       // in the quadric, e.g. a line parallel to and in a cylinder.
-      if (IsInsideQuadric(rayOrigin))
+      // The following condition is a faster equivalent to IsInsideQuadric(rayOrigin)
+      if (Cq <= zero)
```
lesaintjerome, line 130: `Cq<=0` doesn't prevent the source to be ON the quadric (the case `A>0`, `C=0` and `discriminant=0`) and then the ray is tangent. Is the subsequent call to `ApplyClipPlane` still relevant in that case?

src/rtkQuadricShape.cxx
```diff
@@ -98,12 +116,18 @@ QuadricShape ::IsIntersectedByRay(const PointType &  rayOrigin,
...
-    // Check if the central point between the two intersections is in the quadric
-    if (!IsInsideQuadric(rayOrigin + 0.5 * (farDist + nearDist) * rayDirection))
+    // Check if the segment between the two intersections is in the quadric.
+    // Instead of testing if the central point is oustside the quadric with
+    // !IsInsideQuadric(rayOrigin + 0.5 * (farDist + nearDist) * rayDirection),
+    // which was prone to numerical errors, we test if the ray origin is between
+    // intersections (farDist * nearDist < 0) and strictly in the quadric (Cq < 0),
+    // or before / after the two intersections (farDist * nearDist > 0) and outside
+    // the quadric (Cq>0).
+    if (Cq * farDist * nearDist < zero)
```
lesaintjerome: OK. Nice.

lesaintjerome (commented):
I think I would go for a more systematic treatment of the 8 possibles cases of values A, B and C. But I am not enough familiar with the `ApplyClipPlane` function to have a strong opinion.

src/rtkQuadricShape.cxx
SimonRit: The source to intersection(s) distance.

SimonRit: Ok adressed in the next force push

src/rtkQuadricShape.cxx
SimonRit: Yes, it is in my opinion. Consider a cylinder and a ray parallel to the cylinder. Why would it make a difference if the ray is in the cylinder axis (C<0) or on the cylinder (C=0). By definition line 39, we are in the quadric and we can handle both the same way: we have an infinite line which should be clipped.

lesaintjerome: I agree but doesn't the case of a ray ON the cylinder fall in the the previous case (a=b=c=0) rather than this one ? If c=0 and b!=0, the ray is strictly tangent to the quadric at the source point (as in the case a!=0, b=c=0 btw).

SimonRit: Addressed in [last force push](https://github.com/RTKConsortium/RTK/compare/0a0a124f24d5ed813e4c6dff9f430847df6475bb..12c27f990ea4e8022859b901bb307104db0eb848).

lesaintjerome (commented):
Looks good.

PR #823, opened 2025-10-06
DOC: Add CudaCommon documentation

axel-grc:


SimonRit (approved):
I've modified the code to keep using the same logic for downloading file as what is already in place for images.

PR #828, opened 2025-10-13
ENH: Generate rtkConfiguration.h in the sale folder as RTKExport.h

SimonRit:
There is then no need to append the directory to RTK_INCLUDE_DIRS as this is already done by ITK module macros.


axel-grc (commented):
Looks good, but some test did not succeed

PR #836, opened 2025-10-20
ENH: Modify runtime output location for application

axel-grc:


SimonRit (changes requested):
I would personally prefer to replace `${CMAKE_RUNTIME_OUTPUT_DIRECTORY}` by `${RTK_BINARY_DIR}/bin`, I don't see a benefit of creating two new variables.

SimonRit (approved):

PR #839, opened 2025-10-24
 PERF: Improve test duration of examples 

axel-grc:


examples/ConjugateGradient/ConjugateGradient.cxx
```diff
@@ -49,8 +49,8 @@ main(int argc, char * argv[])
   // Create a constant image source for the projections
   auto projectionsSource = rtk::ConstantImageSource<OutputImageType>::New();
   projectionsSource->SetOrigin(itk::MakePoint(-127., -127., -127.));
```
SimonRit: You also need to update the origins to center the projections and the volumes.

SimonRit (commented):
I would center the projections and volumes, as in the original code.

SimonRit (approved):

PR #842, opened 2025-10-29
Updates several documentation pages and adds one

cyrilmory:
Also modifies to way motion compensated 4D conjugate gradient handles DVFs to make it consistent with mcrooster


SimonRit:
Looks good to me @cyrilmory . I have only force pushed a few changes (see diff [here](https://github.com/RTKConsortium/RTK/compare/9bb951cfa5dd6ca2b1845495c3df0da4cb16581b..16896ea78d01f29cf728d8bf520c2b8f9e18341b)) and rebased on main. Should we merge as is or do you want to add more pages to this PR first? I'd prefer to merge it as is unless if you'd prefer to wait.

cyrilmory:
@SimonRit yes, do merge it as is, I will submit another PR for additionnal doc pages

PR #843, opened 2025-10-29
ENH: Remove clang format linter workflow

axel-grc:
We can remove this workflow as the pre-commit workflow handles clang format.


acoussat:
Thanks! If there is no problem with this PR on RTK, can I (or can you) also do it for PCT?

SimonRit (approved):

PR #850, opened 2025-11-07
ENH: Update step size to use minimum spacing if not set

### LINKED ISSUE #803: CudaForwardProjectionImageFilter should use a smart value for the step size
Currently, the value is fixed to 1 mm
https://github.com/RTKConsortium/RTK/blob/main/include/rtkCudaForwardProjectionImageFilter.hxx#L42
We should consider a smart value by default in RTK v3.

axel-grc:
Close #803

### DIFF
--- a/include/rtkCudaForwardProjectionImageFilter.hxx
+++ b/include/rtkCudaForwardProjectionImageFilter.hxx
@@ -76,6 +76,12 @@ CudaForwardProjectionImageFilter<TInputImage, TOutputImage>::GPUGenerateData()
     spacing[i] = this->GetInput(1)->GetSpacing()[i];
   }
 
+  // Determine step size: if user provided a positive StepSize, use it. Otherwise use min(volume spacing)
+  if (m_StepSize == 0.)
+  {
+    m_StepSize = *std::min_element(spacing, spacing + 3);
+  }
+
   // Cuda convenient format for dimensions
   int projectionSize[3];
   projectionSize[0] = this->GetOutput()->GetBufferedRegion().GetSize()[0];
--- a/include/rtkCudaRayCastBackProjectionImageFilter.h
+++ b/include/rtkCudaRayCastBackProjectionImageFilter.h
@@ -74,7 +74,7 @@ public:
   /** Run-time type information (and related methods). */
   itkOverrideGetNameOfClassMacro(CudaRayCastBackProjectionImageFilter);
 
-  /** Set step size along ray (in mm). Default is 1 mm. */
+  /** Set step size along ray (in mm). Default is the min value of spacing. */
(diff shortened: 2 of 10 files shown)


SimonRit (changes requested):
This is nice but the change should be propagated to the adjoint `rtk::CudaRayCastBackProjectionImageFilter` please. The adjoint test should fail, doesn't it?

include/rtkCudaForwardProjectionImageFilter.hxx
```diff
@@ -76,6 +76,12 @@ CudaForwardProjectionImageFilter<TInputImage, TOutputImage>::GPUGenerateData()
     spacing[i] = this->GetInput(1)->GetSpacing()[i];
   }
 
+  // Determine step size: if user provided a positive StepSize, use it. Otherwise use min(volume spacing)
+  if (m_StepSize == 0.)
+  {
+    m_StepSize = static_cast<double>(std::min(spacing[0], std::min(spacing[1], spacing[2])));
```
SimonRit: For style reasons, I'd prefer
```
m_StepSize = *std::min_element(spacing.begin(), spacing.end());
```

SimonRit (changes requested):
Can you please add a note in the [migration guide](https://docs.openrtk.org/en/latest/documentation/docs/rtk_3_migration_guide.html)?

SimonRit:
Can you please change the default value for command line applications to 0? Thanks!

SimonRit:
> This is nice but the change should be propagated to the adjoint `rtk::CudaRayCastBackProjectionImageFilter` please. The adjoint test should fail, doesn't it?

I don't think we addressed this point. Honestly the adjoint tests should not work with this change if we don't change the matched backprojection so I'll investigate.

SimonRit:
It does fail: https://my.cdash.org/viewTest.php?onlyfailed&buildid=3208428

SimonRit (changes requested):
Can you please do the same in [CudaRayCastBackProjectionImageFilter](https://www.openrtk.org/Doxygen/classrtk_1_1CudaRayCastBackProjectionImageFilter.html)? That should fix the failing tests.

include/rtkCudaRayCastBackProjectionImageFilter.h
```diff
@@ -86,7 +86,7 @@ class RTK_EXPORT CudaRayCastBackProjectionImageFilter
   GPUGenerateData();
 
 private:
-  double m_StepSize{ 1. };
+  double m_StepSize{ 0. };
```
SimonRit, line 89: Can you change the comment above `itkGetConstMacro(StepSize, double);` to match the forward projection please?

SimonRit (approved):
Thanks!

PR #851, opened 2025-11-07
Addnoise app

lesaintjerome:
Add two CL applications to add noise to a set of projections:
- `addpoissonnoise`: add Poisson noise based on the ShotNoise filter (adaptation of AddNoise.cxx example).
- `addgaussiannoise`: simple use of `AdditiveGaussianNoiseImageFilter`.

### DIFF
--- a/applications/CMakeLists.txt
+++ b/applications/CMakeLists.txt
@@ -25,6 +25,7 @@ include(${ITK_USE_FILE})
 
 #-----------------------------------------------------------------------------
 # Executables
+add_subdirectory(rtkaddgaussiannoise)
 add_subdirectory(rtkamsterdamshroud)
 add_subdirectory(rtkbackprojections)
 add_subdirectory(rtkfdk)
--- a/applications/CMakeLists.txt
+++ b/applications/CMakeLists.txt
@@ -26,6 +26,7 @@ include(${ITK_USE_FILE})
 #-----------------------------------------------------------------------------
 # Executables
 add_subdirectory(rtkaddgaussiannoise)
+add_subdirectory(rtkaddpoissonnoise)
 add_subdirectory(rtkamsterdamshroud)
 add_subdirectory(rtkbackprojections)
 add_subdirectory(rtkfdk)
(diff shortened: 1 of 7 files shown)


SimonRit:
Thanks @lesaintjerome. I did not feel the need to add these apps but if you think they are helpful, let's move forward with this PR. I'd suggest
- to merge the two applications in a single `rtkaddnoise` app,
- to add a `README.md` app and a reference in the corresponding example,
- to clean-up code in other app doing the same thing (`rtkdrawgeometricphantom`, `rtkdrawshepploganphantom`, `rtkprojectshepploganphantom`).
- to add the Python version.

The commit messages need to follow ITK's style (add a `ENH:` prefix).
@axel-grc can help doing all this, just confirm you agree with my suggestions first.

lesaintjerome:
Thanks for the feedback. One way to go would be to:
- Amend `rtkprojectshepploganphantom` with an option "gaussian" or "poisson" + relevant parameter. 
- Add a `--noise` option to `rtkprojectgeometricphantom` similar to `rtkprojectshepploganphantom`.
This way, no new app, but the need is fullfiled.

The command line interface could be something like:
`--noise (poisson/gaussian) --noise_paramater (I0 mu_ref/sd)`
or 
`-addpoissonnoise (I mu_ref) --addgaussiannoise (std)

As for `rtkdrawXXX`, nothing to change since adding Poisson noise to phantoms would be irrelevant.

What do you think?

axel-grc:
Thanks @lesaintjerome . I think adding options to the projection apps is the way to go.
For the command line arguments, I would like to have:

- One argument for the noise type `--noise (poisson/gaussian)` 
- One for each parameter ` --I0 --mu_ref --std`.

Do you want me to do a first draft ?

lesaintjerome:
Hi @axel-grc. Thanks. I suggested the second CL interface `-addpoissonnoise (I mu_ref) --addgaussiannoise (std)` for 2 reasons:
1. No risk of mixing type and parameters (e.g. --noisetype poisson --sd 10).
2. I guess it could make sense to add both types of noise (model shot noise with Poisson and add detector's electronic noise ??

@SimonRit may have a stronger opinion on that.
And thanks for the proposition of drafting the solution.

SimonRit:
We just discussed it with @axel-grc and @acoussat. We propose:
- we add a section (C++) / group (Python) with the options to add noise
- we add this section / group to all projection apps: `rtkprojections`, `rtkprojectshepploganphantom`, `rtkprojectgeometricphantom`, `rtkforwardprojections`, any other?
- it's a good point that we may want to add both Poisson and Gaussian noise. We agreed on the following:
```
--gaussian: double, standard deviation for the additive Gaussian noise
--poisson: two double, number of photons without object and scaling value of projections for computing Poisson noise
```
Sounds good to you @lesaintjerome ?

lesaintjerome:
Brilliant!

lesaintjerome:
Can I help in anything?

axel-grc:
I will do a first draft!

SimonRit:
We will implement something like that in the targeted applications:
```
ImageType::Pointer output = SetNoiseFromGgo(filter->GetOutput());
```

SimonRit:
Superseded by #860

PR #852, opened 2025-11-12
4D applications

axel-grc:


SimonRit:
A few minor changes in the commit messages please:
FIX -> BUG
Indicate the faulty commit for the Elekta Synergy geometry app
add -> Add, python -> Python in the last commit

axel-grc:
The frames does not update when using cuda gpu

axel-grc:
> The frames does not update when using cuda gpu

I fixed rtkfourdfdk by using directly CudaImageFromImageFilter to keep the conversion filter alive

SimonRit (approved):
Looks good, we'll be tested when 4D examples are added.

PR #858, opened 2025-11-19
ENH: Update RTK_BINARY_DIR to CMAKE_BINARY_DIR

### LINKED ISSUE #844: Applications tests fail on shared windows build since the change of location for applications
See https://my.cdash.org/viewTest.php?onlyfailed&buildid=3177363. Fail started when #836 was merged. Occurs on Windows when RTK is built in ITK with shared libraries.

axel-grc:
Fix #844


SimonRit:
Can we do the same on PCT please?

PR #860, opened 2025-11-21
Add gaussian and poisson noise options to projection applications

axel-grc:


applications/rtknoise_section.ggo
```diff
@@ -0,0 +1,5 @@
+section "Gaussian noise"
+option "gaussian" - "Gaussian noise parameters: <mean> Noise level and <std> Noise standard deviation"    double  multiple                     no
+
+section "Poisson noise"
+option "poisson" -  "Poisson noise parameters: <I0> Number of impinging photons per pixel and <muref> reference linear attenuation coefficient"  double multiple no
```
lesaintjerome: This seems to make the mu_ref parameter mandatory. Why not. Though we could mention something like "e.g. set 0.01879, which is the attenuation coefficient of water at 75 keV".

lesaintjerome: Same could be said about the mean parameter for gaussian noise. Default=0. would make sense, to me.

lesaintjerome (commented):
I did a few tests. It seems to run fine.
The noise feature in `rtkdrawgeometricphantom` and `rtkprojectshepplogan` has not been removed. Should it be?

applications/rtknoise_section.ggo
axel-grc: The problem is that we can't use different default values for lists with gengetopt, so the same default value will be used for mean and std for example.

We can maybe add theses default values to the implementation part ?

lesaintjerome: Or have slightly more complex Gaussian and Poisson ggo sections, with 2 parameters each?

SimonRit (approved):
I have modified this PR to make it to my taste. Main changes are:
- muref now has a default
- mean of the gaussian has been removed
- gaussian is applied both with Poisson (before the log) and without (without any exp / log)

SimonRit:
@axel-grc and @lesaintjerome, do you mind re-reviewing? And testing the Python version @axel-grc if you don't mind.

applications/rtknoise_group.py
```diff
@@ -0,0 +1,92 @@
...
+        help="Comma-separated pre-log Poisson noise parameters: I0 pre-log normalization and (optional) pre-exp multiplicative factor (default 0.01879)",
+        type=float,
+        nargs="+",
+    )
+    rtknoise_group.add_argument(
+        "--gaussian",
+        help="Additive Gaussian noise standard deviation (pre-log if --poisson given, without exp/log otherwise)",
+        type=float,
+    )
+
```
lesaintjerome, line 23: In my small tests, what I had in mind was: take Poisson noise, then add 10% gaussian noise. So I needed to estimate the pre-log values and set the std accordingly).
We could instead pass a "noise level" (e.g. 10%) and the std would be computed as the `pre_log_mean * noise_level`.
But it may conflict with other use cases where the user knows exactly the level of noise he wants to simulate.

applications/rtknoise_group.py
```diff
@@ -0,0 +1,92 @@
...
+    "AddNoiseFromArgParse",
+]
+
+
+# Mimics rtknoise_section.ggo
+def add_rtknoise_group(parser):
+    rtknoise_group = parser.add_argument_group("Poisson and Gaussian noise")
+    rtknoise_group.add_argument(
+        "--poisson",
+        help="Comma-separated pre-log Poisson noise parameters: I0 pre-log normalization and (optional) pre-exp multiplicative factor (default 0.01879)",
```
lesaintjerome: Should we say what does 0.01879 correspond to?

lesaintjerome:
Did some quick tests. It seems to work fine.

applications/rtknoise_group.py
SimonRit: I find it a bit too specific. Why `pre_log_mean * noise_level`? Gaussian noise combined with Poisson noise is mainly here to simulate electronic noise no? So it should be independent of the number of photons emitted? I'm in favor of leaving it as is.

applications/rtknoise_group.py
SimonRit: Addressed with the[ latest force push](https://github.com/RTKConsortium/RTK/compare/48182a14519b6ee783ab0ad3944d5fe1805c6daa..6c063f47083a348d53dd5ce97d1be4d4ec1404ec).

applications/rtknoise_group.py
lesaintjerome: Sure. You're right ! Forget this comment.

axel-grc:
Looks good to me, the only downside is that with poisson noise arguments, the first comma separated value has a default, but not the second, so we need to enter the two values, even if we want the default for the first.
I don't know if there is a solution for that, and with the second argument being optional, that is not really an issue

PR #861, opened 2025-11-21
BUG: Set SetInputFixedVectorLength* of spectral forward and two-step …

SimonRit:


SimonRit:
@cyrilmory This is fixing a test failing with TBB following your work six months ago:
https://my.cdash.org/index.php?project=RTK

PR #863, opened 2025-11-26
STYLE: Centralize verbose option for c++ applications

axel-grc:


SimonRit (approved):
Thanks!

SimonRit (approved):
I changed it to position the common version option at the top, next to --help

acoussat (approved):
Nice solution, in my opinion better than the alternative that used the regular expression! Also, works great on my machine.

PR #866, opened 2025-11-28
ENH: Fail Python tests on warnings during runtime 

axel-grc:
- Add -W error to treat Python warnings as test failures
- Redirect stderr to /tmp/stderr.log to capture C++ warnings from ITK modules
- Check stderr log for "Warning" messages and fail the build if found

This ensures the CI catches both Python and C++ warnings like unknown parameter from SWIG.

### DIFF
--- a/.github/workflows/build-test-package.yml
+++ b/.github/workflows/build-test-package.yml
@@ -47,4 +47,14 @@ jobs:
         pip install $wheel
 
         pip install pytest matplotlib
-        pytest $GITHUB_WORKSPACE/test/*.py -vv -W "ignore:builtin type swig"
+
+        # Run tests with warning detection.
+        # SWIG < 4.4 triggers the Py_LIMITED_API "builtin type swig…" warning when
+        # Python runs with -Walways; pytest enables that behavior, so we ignore it.
+        stderr_log=$(mktemp)
+        pytest $GITHUB_WORKSPACE/test/*.py -vv -s -W "ignore:builtin type swig" 2> "$stderr_log"
+        if grep -q "Warning" "$stderr_log"; then
+          echo "Warnings found in stderr, failing the build"
+          cat "$stderr_log"
+          exit 1
+        fi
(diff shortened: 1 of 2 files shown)


SimonRit (changes requested):
Looks good but can you document in the commit log the reason for adding all these options? Can you also add a commit introducing a warning to check that it works? I would remove the commit before merging.

axel-grc:
I removed rtkPolynomialGainCorrectionImageFilter from WRAPPER_SUBMODULE_ORDER, this should return a warning and the test should fail.

.github/workflows/build-test-package.yml
```diff
@@ -47,4 +47,9 @@ jobs:
         pip install $wheel
 
         pip install pytest matplotlib
-        pytest $GITHUB_WORKSPACE/test/*.py -vv -W "ignore:builtin type swig"
+        pytest $GITHUB_WORKSPACE/test/*.py -vv -s -W error -W "ignore:builtin type swig" 2> /tmp/stderr.log
```
SimonRit: What's the purpose of `-W "ignore:builtin type swig"`?
I'd suggest to use `mktemp` in place of `/tmp/stderr.log` to avoid problems with, e.g., two process writing on the same file.

axel-grc: Just to ignore some known issues in SWIG (tracked in https://github.com/swig/swig/issues/2881). The warnings appear because SWIG is not setting the module attribute on its builtin types.
This should be fixed when ITK use SWIG 4.4.

I changed /tmp/stderr.log for mktemp

SimonRit (changes requested):
That does not seem to work as the tests pass and `rtkPolynomialGainCorrectionImageFilter` jhas been removed from `WRAPPER_SUBMODULE_ORDER`?

SimonRit (approved):
Thanks, please propagate to ITKCudaCommon

PR #876, opened 2025-12-18
ENH: Require attenuation map for JosephAttenuated projection methods

### LINKED ISSUE #452: Attenuated forward and backprojectors should display an error message when the attenuation map is not set
The current behavior is a seg fault, see #450

axel-grc:
Fix #452


SimonRit (changes requested):
Thanks. There should be an error message at the application level as you suggest. But there should also be an exception raised in `JosephBackAttenuatedProjectionImageFilter` as you had initially suggested.

SimonRit (approved):

PR #877, opened 2025-12-19
Displaced detector

### LINKED ISSUE #480: rtkfdk app returns NaN values for parallel geometry
Steps to reproduce : 
rtksimulatedgeometry -n 720 -a 180 -o g720.xml --sdd 0
rtkprojectshepploganphantom -g g720.xml -o sl720.mha  --dimension 500,3 --spacing 1
rtkfdk -g g720.xml -o bpb.mha -p . -r sl720.mha --dimension 500,1,500 --spacing 1

Add --nodisplaced on flag and it works fine.

### LINKED ISSUE #509: DisplacedDetectorForOffsetFieldOfViewImageFilter discards manually set offset
See https://public.kitware.com/pipermail/rtk-users/2022-September/011229.html

axel-grc:
Fix #480 
Closes #509

### DIFF
--- a/include/rtkDisplacedDetectorForOffsetFieldOfViewImageFilter.hxx
+++ b/include/rtkDisplacedDetectorForOffsetFieldOfViewImageFilter.hxx
@@ -65,6 +65,12 @@ DisplacedDetectorForOffsetFieldOfViewImageFilter<TInputImage, TOutputImage>::Gen
                              << "Consider disabling it by setting m_Disable=true "
                              << "or using the nodisplaced flag of the application you are running");
   }
+  else if (this->GetGeometry()->GetSourceToDetectorDistances().front() == 0.)
+  {
+    itkGenericExceptionMacro(<< "Displaced detector cannot handle parallel geometry. "
+                             << "Consider disabling it by setting m_Disable=true "
+                             << "or using the nodisplaced flag of the application you are running");
+  }
 
   using FOVFilterType = typename rtk::FieldOfViewImageFilter<OutputImageType, OutputImageType>;
   auto fieldofview = FOVFilterType::New();
--- a/include/rtkDisplacedDetectorForOffsetFieldOfViewImageFilter.hxx
+++ b/include/rtkDisplacedDetectorForOffsetFieldOfViewImageFilter.hxx
@@ -72,6 +72,12 @@ DisplacedDetectorForOffsetFieldOfViewImageFilter<TInputImage, TOutputImage>::Gen
                              << "or using the nodisplaced flag of the application you are running");
   }
 
+  if (this->GetOffsetsSet())
+  {
+    itkGenericExceptionMacro(<< "DisplacedDetectorForOffsetFieldOfViewImageFilter cannot use manually set offsets. "
+                             << "Use DisplacedDetectorImageFilter when processing independent projections.");
+  }
+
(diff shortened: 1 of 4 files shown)


include/rtkDisplacedDetectorImageFilter.h
```diff
@@ -116,6 +116,13 @@ class ITK_TEMPLATE_EXPORT DisplacedDetectorImageFilter : public itk::InPlaceImag
   itkGetMacro(InferiorCorner, double);
   itkGetMacro(SuperiorCorner, double);
 
+  /** Returns true when offsets were explicitly provided */
+  bool
+  GetOffsetsSet() const
+  {
+    return m_OffsetsSet;
+  }
+
```
SimonRit, line 121: Please use ITK's GetMacro for booleans

include/rtkDisplacedDetectorForOffsetFieldOfViewImageFilter.hxx
```diff
@@ -67,40 +67,89 @@ DisplacedDetectorForOffsetFieldOfViewImageFilter<TInputImage, TOutputImage>::Gen
...
+      typename itk::Index<3>::IndexValueType index =
+        outputLargestPossibleRegion.GetIndex()[0] - outputLargestPossibleRegion.GetSize()[0];
+      outputLargestPossibleRegion.SetIndex(0, index);
+      outputLargestPossibleRegion.SetSize(0, outputLargestPossibleRegion.GetSize()[0] * 2);
+    }
+    else
+    {
+      this->SetInPlace(false);
+      outputLargestPossibleRegion.SetSize(0, outputLargestPossibleRegion.GetSize()[0] * 2);
+    }
```
SimonRit: I don't think we should do the same as in the base class. I suggest to throw an error message instead of copy pasting code, that's will be simpler.

test/rtkdisplaceddetectorcompoffsettest.cxx
```diff
@@ -96,6 +96,30 @@ main(int, char **)
...
+    using CPUDDFType = rtk::DisplacedDetectorImageFilter<OutputImageType>;
+    auto cpuddf = CPUDDFType::New();
+    cpuddf->SetInput(slp->GetOutput());
+    cpuddf->SetGeometry(geometry);
+    cpuddf->SetOffsets(minOffset, maxOffset);
+    cpuddf->InPlaceOff();
+    TRY_AND_EXIT_ON_ITK_EXCEPTION(cpuddf->Update());
+
+    CheckImageQuality<OutputImageType>(cudaddf->GetOutput(), cpuddf->GetOutput(), 1.e-6, 100, 1.);
+  }
```
SimonRit, line 113: The problem of this test is that it is not an offset FOV. In this case, DisplacedDetectorImageFilter can be used.

SimonRit (changes requested):
I suggest "solving" the problem by an error message instead of copy pasting code. You can check in the test that an error message is thrown, I know there is an ITK macro for this.

SimonRit (approved):
Thanks, I think it will do. We might consider merging `DisplacedDetectorImageFilter` and `DisplacedDetectorForOffsetFieldOfViewImageFilter` in the future but that would require a bit more thinking.

SimonRit:
The failing CI is unrelated.

PR #880, opened 2026-01-06
BUG: Remove cuda InlineReconstructionExample test

### LINKED ISSUE #878: rtkInlineReconstructionExampleTest is failing from time to time
The code of the InlineReconstruction example should be revisited to avoid these flukes in the CI.

axel-grc:
The cuda version of InlineReconstructionExample is tested already on the cuda workflow
Fix #878

### DIFF
--- a/test/CMakeLists.txt
+++ b/test/CMakeLists.txt
@@ -365,9 +365,6 @@ rtk_add_cuda_test(rtkFirstCudaReconstructionExampleTest
 rtk_add_test(rtkInlineReconstructionExampleTest
   ../examples/InlineReconstruction/InlineReconstruction.cxx
 )
-rtk_add_cuda_test(rtkCudaInlineReconstructionExampleTest
-  ../examples/InlineReconstruction/InlineReconstruction.cxx
-)
 rtk_add_test(rtkAddNoiseExampleTest
   ../examples/AddNoise/AddNoise.cxx
 )


SimonRit (approved):

SimonRit:
LGTM but I think it would be more readable to modify the file name in the C++ code depending if `RTK_USE_CUDA` is defined, don't you think so?

axel-grc:
> LGTM but I think it would be more readable to modify the file name in the C++ code depending if `RTK_USE_CUDA` is defined, don't you think so?

I think we should keep the example as it is, for the user, there is no point to have different file names if we use cpu or gpu.

SimonRit:
> > LGTM but I think it would be more readable to modify the file name in the C++ code depending if `RTK_USE_CUDA` is defined, don't you think so?
> 
> I think we should keep the example as it is, for the user, there is no point to have different file names if we use cpu or gpu.

Ok. Can you fix the style? precommit complains.

SimonRit:
Actually, there is another problem
https://github.com/RTKConsortium/RTK/actions/runs/20750789634/job/59660497547?pr=880
But while reviewing the code to understand this problem, I realized that the two tests are exactly the same because the example does not use `USE_CUDA`. So I suggest removing the cuda test and depending on `RTK_USE_CUDA`, the CPU or the GPU code will be tested.

SimonRit (approved):
Thanks!

PR #883, opened 2026-01-08
Examples 3D+time

cyrilmory:
This PR moves the code generating the data for 3D+time tests and examples into a helper function, which can be reused.
The changes are currently applied only to rtkfourdroostertest.cxx, but can be propagated to the other 3D+time tests.
The PR also adds the 4D conjugate gradient C++ example, using the helper function to generate the data.

If it is fine for you, I will add the other examples as additional commits on this same PR.


SimonRit:
Thanks. I'm no so fond of having source files out of the example folder, I prefer to have them self contained. I would prefer to have one example to generate the data, a copy of which could be provided on girder, and then other examples to use the generated data. Would that be doable?

cyrilmory:
Yes, that's doable. And would you keep the helper function for the tests, or get rid of it entirely ?

SimonRit:
> Yes, that's doable. And would you keep the helper function for the tests, or get rid of it entirely ?

You can keep it for the tests if it's useful, we've done it with other codes.

cyrilmory:
Is is better now ? If so, I'll proceed with the other 3D+time examples

SimonRit:
I have rebased and then made several modifications to this PR:
https://github.com/RTKConsortium/RTK/compare/bb76a114c163591771c2300ef18b38554795bb10..9163aab7a33048b22e97914cff522f7a6d3580b5

@axel-grc Can you please take over and
* Check that the test run
* Add images in the documentation
* Translate the examples to Python
? Thanks!

SimonRit (changes requested):
I don't understand why you did not keep the same spirit with a code for generating data and others to execute the reconstruction. I think there's less copy pasting, it's better. And I don't think you should keep the [change you made to test/CMakeLists.txt](https://github.com/RTKConsortium/RTK/compare/95c6923764a217145c3f6615c4eae7dab263a41d..23788f473342ffbb26033ff2237837c0132753d1#diff-33394812ba204689144fd2f80832db83853ba1cb32403edb4e15fe4893e675fd). Finally, `Doc:` in the final commit should be `DOC:`.

SimonRit (approved):

PR #884, opened 2026-01-09
WIP: Implement python probe for verbose mode

axel-grc:


SimonRit:
It would be nice to have this option activable on a local developper compilation of the Python package.

PR #889, opened 2026-01-27
Spectral python applications

axel-grc:


SimonRit (approved):
Looks good to me, we will add tests and possibly make some corrections them.

PR #890, opened 2026-01-28
BUG: Include Test depends in test directory

blowekamp:
Addresses build issue with the ITK CMake interfaces library PR.

This fixes missing module dependencies in the test directory that were causing compilation errors when building with updated ITK CMake infrastructure.

The compilation error due to not being able to find the "itkTestingMacros.h" header when building tests.

InsightSoftwareConsortium/ITK#5721

### DIFF
--- a/test/CMakeLists.txt
+++ b/test/CMakeLists.txt
@@ -17,7 +17,13 @@ configure_file(
 if(NOT ITK_DIR)
   set(ITK_DIR ${ITK_BINARY_DIR}/CMakeTmp)
 endif()
-find_package(ITK REQUIRED COMPONENTS ${ITK_MODULE_RTK_DEPENDS})
+find_package(
+  ITK
+  REQUIRED
+  COMPONENTS
+    ${ITK_MODULE_RTK_DEPENDS}
+    ${RTK_TEST_DEPENDS}
+)
 include(${ITK_USE_FILE})
 
 #-----------------------------------------------------------------------------


SimonRit:
Thanks. I don't fully understand the PR but I'll trust you on this if the CI comes green. Just a question: there is one third-party library in RTK, `lpsolve55` which uses `add_library` instead of `itk_module_add_library`:
https://github.com/RTKConsortium/RTK/blob/main/utilities/lp_solve/CMakeLists.txt#L12
Is it necessary to change this for the RTK compatibility with https://github.com/InsightSoftwareConsortium/ITK/pull/5721?

blowekamp:
> Thanks. I don't fully understand the PR but I'll trust you on this if the CI comes green. Just a question: there is one third-party library in RTK, `lpsolve55` which uses `add_library` instead of `itk_module_add_library`: https://github.com/RTKConsortium/RTK/blob/main/utilities/lp_solve/CMakeLists.txt#L12 Is it necessary to change this for the RTK compatibility with [InsightSoftwareConsortium/ITK#5721](https://github.com/InsightSoftwareConsortium/ITK/pull/5721)?

For the use case of lpsolved, where the library is independent and does not require any of the ITK Module specified dependencies, this usage of just itk_module_target is correct. This creates an ITK::lpsolved55 interface library. 

I am also noting that the ITK::RTKModule target includes the lp_solves includes not the target. 
To use the more target properties based approach:
- use target_include_directories on lpsolve55, and target_defines too with correct PUBLIC/PRIVATE.
- add lpsolve55 to RTK_LIBRARIES variable.
- remove lp_solve includes from the  RTK_INCLUDE_DIRS variable.

You can inspect ITKTargets.cmake in the build directory to see the results.

SimonRit:
> > Thanks. I don't fully understand the PR but I'll trust you on this if the CI comes green. Just a question: there is one third-party library in RTK, `lpsolve55` which uses `add_library` instead of `itk_module_add_library`: [`main`/utilities/lp_solve/CMakeLists.txt#L12](https://github.com/RTKConsortium/RTK/blob/main/utilities/lp_solve/CMakeLists.txt?rgh-link-date=2026-01-28T20%3A42%3A15Z#L12) Is it necessary to change this for the RTK compatibility with [InsightSoftwareConsortium/ITK#5721](https://github.com/InsightSoftwareConsortium/ITK/pull/5721)?
> 
> For the use case of lpsolved, where the library is independent and does not require any of the ITK Module specified dependencies, this usage of just itk_module_target is correct. This creates an ITK::lpsolved55 interface library.
> 
> I am also noting that the ITK::RTKModule target includes the lp_solves includes not the target. To use the more target properties based approach:
> 
>     * use target_include_directories on lpsolve55, and target_defines too with correct PUBLIC/PRIVATE.
> 
>     * add lpsolve55 to RTK_LIBRARIES variable.
> 
>     * remove lp_solve includes from the  RTK_INCLUDE_DIRS variable.
> 
> 
> You can inspect ITKTargets.cmake in the build directory to see the results.

Thanks for your detailed answer. Can you please look into the suggestions @axel-grc?

PR #894, opened 2026-02-03
ENH: Adjust RTK applications and example to modern CMake interface

SimonRit:
See
https://docs.itk.org/en/latest/migration_guides/itk_6_migration_guide.html#migration-for-itk-modules for a detailed explanation.

### DIFF
--- a/applications/CMakeLists.txt
+++ b/applications/CMakeLists.txt
@@ -29,6 +29,15 @@ find_package(ITK REQUIRED COMPONENTS ${ITK_MODULE_RTK_DEPENDS})
 include(itkVersion)
 if("${ITK_VERSION_MAJOR}" VERSION_LESS "6")
   include(${ITK_USE_FILE})
+  set(RTK_APPLICATION_TARGETS ${RTK_LIBRARIES})
+else()
+  itk_generate_factory_registration()
+  set(
+    RTK_APPLICATION_TARGETS
+    ITK::RTKModule
+    ITK::ITKImageIO
+    ITK::ITKFFTImageFilterInit
+  )
 endif()
 
 #-----------------------------------------------------------------------------
--- a/examples/ConjugateGradient/CMakeLists.txt
+++ b/examples/ConjugateGradient/CMakeLists.txt
@@ -4,12 +4,25 @@ cmake_minimum_required(VERSION 3.9.5 FATAL_ERROR)
 project(ConjugateGradient)
 
 # Find ITK with RTK
-find_package(ITK REQUIRED COMPONENTS RTK)
-include(itkVersion)
+find_package(
+  ITK
+  REQUIRED
+  COMPONENTS
+    RTK
+    ITKImageIO
+)
 if("${ITK_VERSION_MAJOR}" VERSION_LESS "6")
   include(${ITK_USE_FILE})
+  set(RTK_EXAMPLE_TARGETS ${ITK_LIBRARIES})
+else()
+  itk_generate_factory_registration()
+  set(
+    RTK_EXAMPLE_TARGETS
+    ${ITK_INTERFACE_LIBRARIES}
+    ITK::ITKImageIO
+  )
 endif()
 
 # Executable(s)
 add_executable(ConjugateGradient ConjugateGradient.cxx)
-target_link_libraries(ConjugateGradient ${ITK_LIBRARIES})
+target_link_libraries(ConjugateGradient ${RTK_EXAMPLE_TARGETS})
--- a/test/CMakeLists.txt
+++ b/test/CMakeLists.txt
@@ -27,6 +27,20 @@ find_package(
 include(itkVersion)
 if("${ITK_VERSION_MAJOR}" VERSION_LESS "6")
   include(${ITK_USE_FILE})
+  set(
+    RTK_TEST_TARGETS
+    ${RTK_LIBRARIES}
+    ${ITK_LIBRARIES}
+  )
+else()
+  itk_generate_factory_registration()
+  set(
+    RTK_TEST_TARGETS
+    ITK::RTKModule
+    ITK::ITKImageIO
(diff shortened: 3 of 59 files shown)


dzenanz (approved):

SimonRit:
This currently does not work, it does not register ITK factories for an unknown reason. When executing `rtkfdk` with an `.mha` output, I get
```
ExceptionObject caught with itk::WriteImage(streamerBP->GetOutput(), args_info.output_arg) in file /home/srit/src/rtk/rtk/applications/rtkfdk/rtkfdk.cxx line 207

itk::ImageFileWriterException (0x60b925e7f650)
Location: "unknown" 
File: /home/srit/src/itk/itk/Modules/IO/ImageBase/include/itkImageFileWriter.hxx
Line: 117
Description:  Could not create IO object for writing file fdk.mha
  Tried to create one of the following:
    HndImageIO
    HncImageIO
    XimImageIO
    HisImageIO
    ImagXImageIO
    DCMImagXImageIO
    EdfImageIO
    XRadImageIO
    OraImageIO
    HndImageIO
    HncImageIO
    XimImageIO
    HisImageIO
    ImagXImageIO
    DCMImagXImageIO
    EdfImageIO
    XRadImageIO
    OraImageIO
    GDCMImageIO
  You probably failed to set a file suffix, or
    set the suffix to an unsupported type.
```
If I rollback to using `include(${ITK_USE_FILE})` in `applications/CMakeLists.txt`, it does work so there must be something in `UseITK.cmake` missing in my code. Any pointer is welcome @blowekamp.

SimonRit:
In the latest force push, I added [this code](https://github.com/RTKConsortium/RTK/pull/894/files#diff-22b5c767b274cb860bc3604b376958ddc1bc588a298153c14dcaf6d95db676b1R34-R55) copy pasted from UseITK.cmake. Then, RTK's applications work. However, the examples (external applications using RTK) still fail due to missing factories so something's not right and I don't know what.

examples/ConjugateGradient/CMakeLists.txt
```diff
@@ -5,11 +5,16 @@ project(ConjugateGradient)
...
+  itk_generate_factory_registration()
 endif()
 
 # Executable(s)
 add_executable(ConjugateGradient ConjugateGradient.cxx)
-target_link_libraries(ConjugateGradient ${ITK_LIBRARIES})
+if("${ITK_VERSION_MAJOR}" VERSION_LESS "6")
+  target_link_libraries(ConjugateGradient ${ITK_LIBRARIES})
+else()
+  target_link_libraries(ConjugateGradient ITK::RTKModule)
```
blowekamp: You were previously linking to all loaded ITK modules, and presumably using them with "UseITK".

Using `ITK_INTERFACE_LIBRARIES` will apply properties and line to all the `ITK::{module-name}Module` loaded in the find_package. This may include additional modules no specified as dependents of just the RTK Module.

applications/CMakeLists.txt
```diff
@@ -29,6 +29,30 @@ find_package(ITK REQUIRED COMPONENTS ${ITK_MODULE_RTK_DEPENDS})
...
+    string(TOUPPER ${_factory_name} _factory_uc)
+
+    if(NOT ITK_NO_${_factory_uc}_FACTORY_REGISTER_MANAGER)
+      set_property(
+        DIRECTORY
+        APPEND
+        PROPERTY
+          COMPILE_DEFINITIONS
+            ITK_${_factory_uc}_FACTORY_REGISTER_MANAGER
+      )
```
blowekamp: The migration guide contains some information on "Factory Registration with Meta Modules": https://docs.itk.org/en/latest/migration_guides/itk_6_migration_guide.html#application-configuration

The docs are missing that an application should also link to the meta-module "ITK::ITKImageIO". These meta-module have the properties of the include and compile definitions in them so they don't need to be set in at a directory/global level.

blowekamp (commented):
Thank you for trying out the latest changed in ITK right away.

You may want to create a linking function or some variables to avoid the duplicated conditionings.

I was uncertain what to run to reproduce your error messages.

applications/CMakeLists.txt
SimonRit: Thanks, I have just force-pushed a new version which works. I am not entirely convinced though. Can I make the target `RTKModule` dependent on `ITK::ITKImageIO` and `ITK::ITKFFTImageFilterInit` to avoid having to specify it in each application and example?

By the way, the doc mentions `ITKIOFFT` which does not exist, I have used `ITKFFTImageFilterInit` instead.

Thanks again for you support!

SimonRit:
@blowekamp I'm ready to merge this PR based on your documentation. If you've time to quickly check, that would be great. No pb if you don't.

test/CMakeLists.txt
```diff
@@ -27,18 +27,28 @@ find_package(
...
+  )
+else()
+  itk_generate_factory_registration()
+  set(
+    RTK_TEST_TARGETS
+    ITK::RTKModule
+    ITK::ITKImageIO
+    ITK::ITKFFTImageFilterInit
+    ITK::ITKTestKernel
+  )
```
blowekamp, line 43: This is non-standard code for an ITK modular test directory.

The `itk_module_test` provides a `RTK-Test_LIBRARIES` variable populated with what is specified in the itk_module `TEST_DEPENDS` options. This LIBRARIES variable is then passed to create test driver as the second argument or passes as a target_link_libraries.

But maybe you are trying to support building this directory as a stand alone directory or some alternate configuraiton.

blowekamp:
Looks good to me.

test/CMakeLists.txt
SimonRit: This is a separate issue but there is no good reason for this. We'll work on that with @axel-grc, tracked in #896.

PR #899, opened 2026-02-11
COMP: Fix cmake option for RTK_CUDA_VERSION

### LINKED ISSUE #897: CMake option RTK_CUDA_VERSION should not be a boolean
The work done in https://github.com/RTKConsortium/ITKCudaCommon/commit/7d446f45735862d40a7d796a7820041a7aec01d3 should be propagated to RTK.

axel-grc:
Fix #897

### DIFF
--- a/itk-module-init.cmake
+++ b/itk-module-init.cmake
@@ -49,18 +49,31 @@ if(CMAKE_CUDA_COMPILER)
 endif()
 option(RTK_USE_CUDA "Use CUDA for RTK" ${RTK_USE_CUDA_DEFAULT})
 
-# Configure CUDA compilation options
-option(
-  RTK_CUDA_VERSION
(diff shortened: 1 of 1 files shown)


itk-module-init.cmake
```diff
@@ -50,17 +50,29 @@ endif()
...
+  )
+else()
+  set(
+    RTK_CUDA_VERSION
+    ${RTK_CUDA_VERSION}
+    CACHE STRING
+    "Specify the exact CUDA version that must be used for RTK"
+  )
+endif()
+mark_as_advanced(RTK_CUDA_VERSION)
```
SimonRit: This should be in the `if(RTK_USE_CUDA)` condition I believe

SimonRit (commented):
Only one inline suggestion before merging

SimonRit (approved):

PR #900, opened 2026-02-12
ENH: Refactor test/CMakeLists.txt to adhere to ITK module convention

### LINKED ISSUE #896: Refactor test/CMakeLists.txt to adhere to ITK module convention
This requires removing the compilation of executables and creating a test driver instead. See RTKConsortium/ITKCudaCommon for an exemple. Issue generated following https://github.com/RTKConsortium/RTK/pull/894#discussion_r2764145519.

axel-grc:
- Changed the main function in various test files to specific test function names .
- Updated the CMakeLists of tests to follow itk convention with createtestdriver

Close #896

### DIFF
--- a/cmake/KWStyle/RTKOverwrite.txt
+++ b/cmake/KWStyle/RTKOverwrite.txt
@@ -1,4 +1,6 @@
 test/.*\.cxx Namespace Disable
 test/.*\.cxx Header Disable
+test/.*\.cxx Comments Disable
+test/.*\.h Namespace Disable
 examples/.*\.cxx Namespace Disable
 examples/.*\.cxx Header Disable
--- a/examples/AddNoise/AddNoise.cxx
+++ b/examples/AddNoise/AddNoise.cxx
@@ -15,7 +15,7 @@ const float I0 = 1e4;     // Number of photons before attenuation
 const float mu = 0.01879; // mm^-1, attenuation coefficient of water at 75 keV
 
 int
-main()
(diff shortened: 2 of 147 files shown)


axel-grc:
There is a problem with the example tests: with the ITK convention, tests are dispatched by the test driver and must provide an entrypoint like int <TestName>(int,char**). The examples are standalone programs with a main(), and I don’t think it’s a good idea to rewrite their main just to fit the test driver pattern.

For now, I kept the previous approach only for examples: build each example as its own executable and register it with CTest via itkTestDriver.

Do you have any insight on this, @SimonRit

SimonRit:
> There is a problem with the example tests: with the ITK convention, tests are dispatched by the test driver and must provide an entrypoint like int (int,char**). The examples are standalone programs with a main(), and I don’t think it’s a good idea to rewrite their main just to fit the test driver pattern.
> 
> For now, I kept the previous approach only for examples: build each example as its own executable and register it with CTest via itkTestDriver.
> 
> Do you have any insight on this, @SimonRit

I'm not following. This seems to be the convention of a `main` function in C which is ok to adopt for all examples as is done in `FirstReconstruction`: 
https://github.com/RTKConsortium/RTK/blob/main/examples/FirstReconstruction/FirstCudaReconstruction.cxx#L12

SimonRit (approved):
Thanks!

SimonRit (approved):

PR #901, opened 2026-02-12
ENH: Refactor lpsolve CMake configuration for target-oriented usage

### LINKED ISSUE #891: Revisit lpsolve CMake code with a target-oriented approach
See https://github.com/RTKConsortium/RTK/pull/890#issuecomment-3813834068

axel-grc:
Close #891

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -172,17 +172,9 @@ endif()
 #=========================================================
 add_subdirectory(utilities/lp_solve)
 
-set(
-  LPSOLVE_INCLUDE_DIRS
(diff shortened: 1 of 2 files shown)


SimonRit (approved):
Looks good to me. Have you tested that it works:
- for RTK build along ITK, using the build and the install dir
- for RTK built independently?

axel-grc:
Yes it works for both @SimonRit 

> Looks good to me. Have you tested that it works:
> 
>     * for RTK build along ITK, using the build and the install dir
> 
>     * for RTK built independently?

PR #905, opened 2026-02-25
Add rtkvarianprobeamgeometry python app

axel-grc:


SimonRit:
If InsightSoftwareConsortium/ITK#5833 is merged and an ITK release prepared soon after, this will need to be modified.

SimonRit:
@axel-grc, can you test with the new ITK 5.4.6?

.github/workflows/build-test-package-python-cuda.yml
```diff
@@ -3,7 +3,7 @@ name: 'Package RTK with CUDA'
 on: [push,pull_request]
 
 env:
-  itk-wheel-tag: 'v5.4.5'
+  itk-wheel-tag: 'v5.4.6'
```
SimonRit: Please drop this commit and test without it first. It's not necessary for this PR I believe.

SimonRit (approved):

PR #906, opened 2026-02-26
ENH: Update CUDA image filter wrappers to follow itkcudacommon

axel-grc:


SimonRit:
Waiting for CudaCommon release to relaunch the tests. If they are then green, will merge

SimonRit:
The second commit will have to be dropped when #885 is merged to main

SimonRit (approved):
Thanks

PR #913, opened 2026-03-03
 ENH: Add optional and groups dependencies in pyproject.toml 

### LINKED ISSUE #908: Use optional-dependencies in pyproject.toml for optional Python dependencies
See https://paddyroddy.github.io/talks/python-packaging/#/2/9. We could add a `gui` one installing `pyvista`. We may also consider `dependency-groups` for development.

axel-grc:
Close #908

### DIFF
--- a/GettingStarted.md
+++ b/GettingStarted.md
@@ -1,5 +1,4 @@
-Getting started
-===============
+# Getting started
 
 Here are suggested steps for the RTK beginner.
 
--- a/INSTALLATION.md
+++ b/INSTALLATION.md
@@ -1,8 +1,6 @@
-RTK installation
-================
+# RTK installation
 
-Configuration, compilation and installation with ITK
-----------------------------------------------------
+## Configuration, compilation and installation with ITK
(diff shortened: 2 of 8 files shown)


SimonRit (changes requested):
Thanks! My understanding is that the `dev` one should be in [dependency-groups](https://packaging.python.org/en/latest/specifications/dependency-groups/) and split into `docs` and `test`. Any reason for not doing this?

axel-grc:
> Thanks! My understanding is that the `dev` one should be in [dependency-groups](https://packaging.python.org/en/latest/specifications/dependency-groups/) and split into `docs` and `test`. Any reason for not doing this?

Not really, I will change that.

axel-grc:
From what I understand, [dependency-groups](https://packaging.python.org/en/latest/specifications/dependency-groups/) is the newer standard and replace optional-dependencies.
So I used dependency-groups for gui too

SimonRit:
> From what I understand, [dependency-groups](https://packaging.python.org/en/latest/specifications/dependency-groups/) is the newer standard and replace optional-dependencies. So I used dependency-groups for gui too

This is not my understanding: https://pydevtools.com/handbook/explanation/what-are-optional-dependencies-and-dependency-groups/. Any resource for justifying this?

axel-grc:
Ok I didn't get the differences, I will change this again

SimonRit:
It would be good to document (in INSTALLATION.md and rtkshowgeometry page) and test before merging.

SimonRit (changes requested):
When pyvista is not installed, I get the message `ImportError: rtkshowgeometry requires the 'pyvista' package. Please install it with 'pip install pyvista' to use this feature.` Can we change this according to the new `pip install itk-rtk[gui]` please?

SimonRit (approved):

PR #914, opened 2026-03-04
Fix typo in rtk3Doutputimage_group.py

acoussat:
When `--size` is provided, the variable is set to `args_info.size` and not `args_info.Size`, as added by #833. This fix avoids `AttributeError`s when the check is triggered.

```
AttributeError: 'Namespace' object has no attribute 'Size'. Did you mean: 'size'?
```


SimonRit (approved):
Thanks

PR #916, opened 2026-03-05
COMP: Drop use of itk::Math::Absolute to fix compilation warnings

SimonRit:
InsightSoftwareConsortium/ITK#5802 and #904 have triggered new warnings such as
```
[CTest: warning matched] /home/srit/src/rtk/dashboard_tests/RTK/include/rtkJosephBackProjectionImageFilter.hxx:277:28: warning: comparison of integer expressions of different signedness: 'int' and 'unsigned int' [-Wsign-compare]
  277 |         for (int i{ 0 }; i < itk::Math::Absolute(fs - ns) - 1; ++i)
```
The new behavior is intentional but returning an unsigned is problematic in this case as we subtract 1 from it. So we now use `std::abs` throughout the code

### DIFF
--- a/applications/rtkprojectionmatrix/rtkprojectionmatrix.cxx
+++ b/applications/rtkprojectionmatrix/rtkprojectionmatrix.cxx
@@ -114,7 +114,7 @@ main(int argc, char * argv[])
     return EXIT_FAILURE;
   }
   const OutputImageType::DirectionType dir = reader->GetOutput()->GetDirection();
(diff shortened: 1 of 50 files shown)


SimonRit:
See discourse discussion https://discourse.itk.org/t/how-to-best-handle-itk-absolute/7724 on this topic.

SimonRit:
Failing test is due to recent changes in CudaCommon, will be fixed with #906 when CudaCommon is released (soon)

PR #918, opened 2026-03-10
COMP: Conditionally restore legacy lpsolve include behavior for ITK<6

axel-grc:


SimonRit (commented):
That does not really look like a solution but I don't really understand the fix, do you think it is better than reverting 967c0e976558ba5843caf84ff1244337dd2563d0?

axel-grc:
> That does not really look like a solution but I don't really understand the fix, do you think it is better than reverting [967c0e9](https://github.com/RTKConsortium/RTK/commit/967c0e976558ba5843caf84ff1244337dd2563d0)?

I think it is better to find a fix to follow modern convention and use target-oriented path

SimonRit (changes requested):
The proposed change comes down to removing any lpsolve include from the public interface. In this case, we won't have an include problem but that does not explain why there was an include problem. Why not though but in this case, we should stop installing the lpsolve headers. 
The change from y to z is not good in my opinion, lpsolve is solving a 2D problem and I'd like to stick to x y even if in FieldOfViewImageFilter we use it in a y slice.

SimonRit:
> > That does not really look like a solution but I don't really understand the fix, do you think it is better than reverting [967c0e9](https://github.com/RTKConsortium/RTK/commit/967c0e976558ba5843caf84ff1244337dd2563d0)?
> 
> I think it is better to find a fix to follow modern convention and use target-oriented path

Quickly looking at what happened, the modern convention is only handled in ITK 6. If we want to keep supporting the latest release, we should conditionally use the old way for ITK_VERSION_MAJOR<6. See
https://github.com/InsightSoftwareConsortium/ITK/commit/87cc629cd4ee799af9f41b01d52cb6b894f0aa59
Rather than completly modifying our handling of lp_solve, can we try to mix the old and the new way? Basically revert #901 for ITK_VERSION_MAJOR 5 or "apply" https://github.com/InsightSoftwareConsortium/ITK/commit/87cc629cd4ee799af9f41b01d52cb6b894f0aa59 on the RTK side for ITK_VERSION_MAJOR.

SimonRit (approved):
Thanks for the temp patch

PR #926, opened 2026-04-01
Improve CI

### LINKED ISSUE #924: Implement CI Python tests on MacOS and Windows
Similarly to what's done on Linux, it would be good if the Python packages were tested on MacOS and Windows (with and without for the latter).

### LINKED ISSUE #938: Python tests should not run on self-hosted for non-CUDA packages

axel-grc:
Close #924
Close #938


.github/workflows/build-test-package-python-cuda.yml
```diff
@@ -267,3 +267,41 @@ jobs:
           echo "$matches"
           exit 1
         fi
+
+  test-python-wheel-windows:
```
SimonRit: Can't we merge the two test-python-wheel workflows? They seem very similar to me, it would be nicer to avoid all this redundant code

SimonRit (changes requested):
Can we avoid redundant workflows? To be discussed.

SimonRit (approved):
Nice work!

PR #927, opened 2026-04-01
Python app

axel-grc:


SimonRit:
It would be nice to add some documentation examples for each application

PR #928, opened 2026-04-09
Make application argument names consistent

### LINKED ISSUE #923: Mismatched options between C++ and Python version
The C++ options `--output-every` and  `--iteration-file-name` became `--outputevery` and  `--iterationfilename` in Python applications for iterative reconstruction. Was this intentional? I would prefer to have matched options.

axel-grc:
Close #923 

I preferred harmonize all application argument names by using hyphens according to gnu convention

### DIFF
--- a/CodeContribution.md
+++ b/CodeContribution.md
@@ -4,6 +4,8 @@
 
 RTK is based on ITK and aims at following its coding conventions. Any developer should follow these conventions when submitting new code or contributions to the existing one. We strongly recommend you to read thoroughly [ITK's style guide](http://www.itk.org/Wiki/ITK/Coding_Style_Guide).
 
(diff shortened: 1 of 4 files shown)


SimonRit (changes requested):
I'm sorry, as suggested in https://github.com/RTKConsortium/RTK/issues/923#issuecomment-4161511571, this is a too brutal backward incompatible change. I would only accept it if we keep maintaining the old options for a while but I also have the feeling this is heavy and unnecessary. To be discussed.

SimonRit (changes requested):
Thanks, looks good. Can you please add a note somewhere (maybe in https://docs.openrtk.org/en/latest/applications/README.html) on what's the preferred style for options? It can be in a new commit I think.

SimonRit (approved):
Thanks

PR #930, opened 2026-04-13
COMP :  Update to the new cuCtxCreate() for cuda 13

axel-grc:
CUDA 13.0 has replaced the older cuCtxCreate() function with a new version, so th  CUDA 13 cuCtxCreate signature is used with CUctxCreateParams in GetFreeGPUGlobalMemory, while keeping a fallback for older CUDA versions via CUDA_VERSION guards.


SimonRit (approved):
I made one additional change for the test-python-wheel workflow and that fixed it. Thanks!

PR #932, opened 2026-04-15
BUG: Link FFT factory meta-module in test drivers for ITK 6

axel-grc:


SimonRit:
Fixed by InsightSoftwareConsortium/ITK#6073

PR #933, opened 2026-04-17
fix: Python interface compatibility - update requires-python and fix examples

ethanio12345:
## Description

This PR fixes the Python interface for the RTK package to work with the current itk-rtk PyPI distribution.

### pyproject.toml Fix
- Updated `requires-python` from ">=3.8" to ">=3.10" to resolve dependency conflicts with `sphinx>=7.2.6` and `myst-parser[linkify]`
- Added `uv.lock` for dependency reproducibility

### Python Example Fixes
- **FirstReconstruction.py**: Fixed `FDKConeBeamReconstructionFilter` template parameter count (3→1)
- **InlineReconstruction.py**: Fixed `ParkerShortScanImageFilter` and `FDKConeBeamReconstructionFilter` template parameter counts
- **ConjugateGradient.py**: Rewritten to use core RTK API instead of application-level functions not in PyPI package
- **GeometricPhantom.py**: Switched to built-in Shepp-Logan phantom instead of missing "Thorax" config file
- **WaterPreCorrection.py**: Added auto-generation of test data when output directory is missing

### Other
- Updated `.gitignore` to exclude generated example output files

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -33,3 +33,9 @@ CMakeLists.txt.user*
 
 # Mac System File
 .DS_Store
+build/
+
+# Generated output files from examples
+*.mha
+geometry.xml
+output.xml
--- a/examples/ConjugateGradient/ConjugateGradient.py
+++ b/examples/ConjugateGradient/ConjugateGradient.py
@@ -12,26 +12,27 @@ sid = 600
 sdd = 1200
 scale = 2.0
 
-# Using RTK applications python APIs to create the geometry and projections
-rtk.rtksimulatedgeometry(
-    nproj=numberOfProjections,
-    arc=angularArc,
-    sid=sid,
-    sdd=sdd,
-    output="geometry.xml",
-)
-rtk.rtkprojectgeometricphantom(
-    geometry="geometry.xml",
-    output="projections.mha",
-    phantomfile="Thorax",
-    phantomscale=scale,
-    rotation="1,0,0,0,0,-1,0,1,0",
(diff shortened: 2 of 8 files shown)


ethanio12345:
Actually, I might leave these open (at least for y'all to close). I had some package conflicts and uv suggests py 3.10 for the sphinx. There were some missing files not available when I did git clone to run the examples.

SimonRit:
I think the issue is that the current package release is incompatible with the doc on the main branch since #919. However, you can download the latest python package as GitHub artifacts on the CI as artifacts from the workflows https://github.com/RTKConsortium/RTK/pull/919/checks.

@axel-grc should we consider rolling back the doc to one compatible with the current pypi packages and produce python packages with backward compatibility until the next release?

acoussat:
As a more general comment, shouldn't the landing page of https://docs.openrtk.org be https://docs.openrtk.org/en/v2.7.0/ instead of https://docs.openrtk.org/en/latest/?

axel-grc:
> instead

That could be a solution, but that would hide the latest changes in the c++ side.
Maybe we could consider to add a warning in the latest doc to refer to the 2.7 doc when installing the package with pip install itk-rtk ?

SimonRit:
I prefer to use latest as the default. +1 for adding a note in INSTALLATION.md on how to download the latest package.

SimonRit:
Superseded by the PR #934 updating the documentation.

PR #935, opened 2026-04-22
ENH: Convert ExternalData .md5 tags to .cid (IPFS content IDs)

hjmjohnson:
Migrate ExternalData content-link stubs from legacy md5/sha512 hashes to IPFS .cid tags, aligning with the ITK-wide CID migration.


SimonRit (changes requested):
Thanks for moving the remote module along with the ITK main repository. I haven't followed the reason for this change but I'd The PR does not pass the CI. Can you have a look please?

SimonRit (approved):

SimonRit:
Thanks @hjmjohnson, it works now!

PR #936, opened 2026-04-23
Fix pytest command in documentation

acoussat:
The command currently given in the documentation is wrong and does not run any test.


axel-grc:
Thanks @acoussat  !

PR #940, opened 2026-05-12
COMP: Remove unused variable declarations

axel-grc:


SimonRit (changes requested):
I'm not sure what you mean with REF: in the commit message. Do you mean STYLE:?

axel-grc:
> I'm not sure what you mean with REF: in the commit message. Do you mean STYLE:?

We can use COMP right ?

SimonRit (approved):

PR #941, opened 2026-05-12
ENH: Fix CUDA 2D FFT convolution for multi-projection inputs

### LINKED ISSUE #557: ITK-master/Modules/Remote/RTK/src/rtkCudaFFTProjectionsConvolutionImageFilter.cu:83: itk::ERROR: CUDA ERROR: out of memory
I used "CudaScatterGlareCorrectionImageFilter" to do some scatter glare correction test, but with a projection file,  I met the error as title described. the projection size is 768*768*396, when i used a projection file which size is 768*768*132, it worked fine!
by the way when error took place, the GPU memory was sufficent!
image

axel-grc:
Process multi-projection 2D-kernel convolutions slice-by-slice in CudaFFTProjectionsConvolutionImageFilter instead of building one large stack-wide FFT problem.

This avoids the oversized padded convolution volume seen with CudaScatterGlareCorrectionImageFilter on large projection stacks

Fix #557

### DIFF
--- a/include/rtkCudaFFTProjectionsConvolutionImageFilter.hxx
+++ b/include/rtkCudaFFTProjectionsConvolutionImageFilter.hxx
@@ -91,85 +91,138 @@ template <class TParentImageFilter>
 void
 CudaFFTProjectionsConvolutionImageFilter<TParentImageFilter>::GPUGenerateData()
 {
-  // Pad image region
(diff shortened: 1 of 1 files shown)


SimonRit:
Thanks for the fix. Projection-by-projection processing should be handled a priori, not in the filter in my opinion as this might have a cost computing time wise.

PR #946, opened 2026-05-18
ENH: Cleanup "ifdef CudaCommon_VERSION_MAJOR"

### LINKED ISSUE #944: Cleanup "ifdef CudaCommon_VERSION_MAJOR"
We are only compatible with the latest CudaCommon release and the HEAD, this is always defined since CudaCommon v2.

axel-grc:
Fix #944

### DIFF
--- a/src/rtkXRadImageIOFactory.cxx
+++ b/src/rtkXRadImageIOFactory.cxx
@@ -18,8 +18,11 @@
 
 #include "rtkXRadImageIOFactory.h"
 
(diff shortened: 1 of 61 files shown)


src/rtkXRadImageIOFactory.cxx
SimonRit: Can you try to remove the itk namespaces? I don't think we need them

SimonRit (approved):

Issue #943, opened 2026-05-13
Convert remaining sha512 data file to cid and document use


SimonRit:

SimonRit:
To be done in RTKConsortium/PCT too.

SimonRit:
All testing data are using CID, we keep the doc with ITK's Girder yet, seems too complicated to use CIDs.

PR #949, opened 2026-05-20
ENH: Update documentation and code to use CID instead of remaining SH…

### LINKED ISSUE #948: Doc images are no longer downloaded
This is probably linked to the conversion from sha512 to cid files and the CMake script `copy_and_fetch_sphinx_doc_files.cmake` not being updated.

axel-grc:
Fix #948


SimonRit:
Can you also fix #943 in this PR please?

PR #960, opened 2026-06-18
COMP: Fix format string

axel-grc:
Replace the snprintf-based iteration output formatting with a direct substitution of the documented %d placeholder.

This avoids GCC's -Wformat-nonliteral warning when the filename pattern comes from --iteration-file-name, while preserving the supported behavior.


SimonRit (approved):
Thanks for this. The failing CI is unrelated I believe.
When the minimum C++ standard wil be 20, we'll switch to using `std::format`.

PR #966, opened 2026-07-10
ENH: Support CUDA ray cast projectors with parallel geometry

### LINKED ISSUE #929: Implement CudaRayCast for parallel geometries
The CudaRayCast forward and back projector currently raise an exception
https://github.com/RTKConsortium/RTK/blob/main/src/rtkCudaRayCastBackProjectionImageFilter.cxx#L37
https://github.com/RTKConsortium/RTK/blob/main/src/rtkCudaWarpForwardProjectionImageFilter.cxx#L136
It should not be too hard to make them compatible with parallel geometries

axel-grc:
Remove the parallel-geometry rejection from CUDA ray cast forward, warped forward, and back projectors. Add parallel ray construction using the geometry projection direction, and cover it in the CUDA ray-cast adjoint test.
Also align the source-position buffer type and size with the CUDA float3 layout, removing an unsafe double-pointer cast.

Close #929


SimonRit:
Please add a large test for cone-beam to see if this modification has a computational cost. If it does, move the condition out of the kernel (make two kernels).

SimonRit:
Please add a test in rtkforwardprojectiontest.cxx for the parallel geometry and flat panel detector

axel-grc:
I made some test on cone beam before and after this PR, and it seems there is no computational cost @SimonRit

PR #967, opened 2026-07-10
ENH: Support install-tree use of RTK when built externally

### LINKED ISSUE #957: Fix installation of wrapping when compilation is done separately from ITK
RTK currently prevents wrapping installation when compilation is not done with ITK's:
https://github.com/RTKConsortium/RTK/blob/main/CMakeLists.txt#L355
According to a [discussion on PCT](https://github.com/RTKConsortium/PCT/pull/69/changes#r2816045672), this should be possible.

axel-grc:
After some tests, it seems  possible to install standalone modules, so I removed the error message.

Close #957

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -210,26 +210,7 @@ endif()
 "
 )
 
-set(
-  RTK_EXPORT_CODE_INSTALL
-  "
-# The RTK version number
-set(RTK_VERSION_MAJOR ${RTK_VERSION_MAJOR})
-set(RTK_VERSION_MINOR ${RTK_VERSION_MINOR})
-set(RTK_VERSION_PATCH ${RTK_VERSION_PATCH})
-
-# Whether the compiled version of RTK uses CUDA
-set(RTK_USE_CUDA ${RTK_USE_CUDA})
-
-# Whether RTK applications were built
-set(RTK_BUILD_APPLICATIONS ${RTK_BUILD_APPLICATIONS})
-
-if(RTK_USE_CUDA)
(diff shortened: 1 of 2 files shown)


SimonRit (changes requested):
That doesn't work for me. If I compile and install ITK without RTK, then compile and install RTK, both with the same `CMAKE_INSTALL_PREFIX`, then try to compile the `FirstReconstruction` example againt this install directory, I get: 
```
  Target "FirstReconstruction" links to:

    ITK::RTKModule

  but the target was not found.  Possible reasons include:

    * There is a typo in the target name.
    * A find_package call is missing for an IMPORTED target.
    * An ALIAS target is missing.



CMake Error at CMakeLists.txt:34 (target_link_libraries):
  Target "FirstCudaReconstruction" links to:

    ITK::RTKModule

  but the target was not found.  Possible reasons include:

    * There is a typo in the target name.
    * A find_package call is missing for an IMPORTED target.
    * An ALIAS target is missing.
```
On Ubuntu with ITK main and CMake 3.29. Don't you have the same issue? This is precisely what motivated this check. BTW, I had to implement some fixes for lpsolve with other configurations, see #976.

axel-grc:
I don't know why it worked on my side, I maybe not tested it on a fresh environment, but you are right.
but now it should work.
The install(EXPORT ...) in ITKModuleExternal.cmake doesn't generate install code for externally-built modules. The fix adds an explicit install(EXPORT ...) to Modules/Targets/

SimonRit (approved):
Thanks!

PR #973, opened 2026-07-20
ENH: Replace ExtractImageFilter with zero-copy sub-region extraction

### LINKED ISSUE #91: Time is lost extracting each (substack of) projection(s) in the mini-pipeline
Creating the (substack of) projection(s) from the pointer would be sufficient. CPU and GPU versions would be required.

axel-grc:
Fix #91 : Replace itk::ExtractImageFilter with a lightweight free function that creates sub-region views by sharing the input buffer via SetImportPointer, avoiding pixel copies and filter pipeline overhead.

### DIFF
--- a/include/rtkFDKConeBeamReconstructionFilter.h
+++ b/include/rtkFDKConeBeamReconstructionFilter.h
@@ -23,8 +23,7 @@
 #include "rtkConfiguration.h"
 #include "rtkFDKBackProjectionImageFilter.h"
 #include "rtkFFTRampImageFilter.h"
-
-#include <itkExtractImageFilter.h>
+#include "rtkSubRegionViewImageFilter.h"
 
 namespace rtk
 {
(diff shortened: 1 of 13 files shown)


axel-grc:
I made some tests on FDK, the extract phase is at least 1000x faster, but it is still a very small part of the total reconstruction time (<1%).

SimonRit (changes requested):
I don't think this works. The proposed solution only works if the extracted region corresponds to a contiguous memory region of the input. This is not always the case and there is a check missing for this. If it is not contiguous, there should be a fallback to `itk::ExtractImageFilter`.

include/rtkExtractImageSubRegion.h
```diff
@@ -0,0 +1,126 @@
...
+    using TCudaImage = itk::CudaImage<PixelType, Dimension>;
+    if (TCudaImage * cudaOutput = dynamic_cast<TCudaImage *>(output.GetPointer()))
+    {
+      cudaOutput->GetModifiableDataManager()->SetBufferSize(numPixels * sizeof(PixelType));
+      cudaOutput->GetModifiableDataManager()->SetImagePointer(cudaOutput);
+      cudaOutput->GetModifiableDataManager()->SetCPUBufferPointer(const_cast<PixelType *>(bufferPtr));
+      cudaOutput->GetModifiableDataManager()->SetGPUDirtyFlag(true);
+      cudaOutput->GetModifiableDataManager()->SetCPUDirtyFlag(false);
+    }
+#endif
```
SimonRit: I don't think CudaImage needs a specific case when dealing with CPU memory

SimonRit:
The continuous memory test should be updated and the code should be wrapped in an image filter if possible.

PR #974, opened 2026-07-20
ENH: Fix Parker weighting to not discard first and last projections

### LINKED ISSUE #870: Parker weighting should not throw away the first and the last projections
In RTK, we currently assume that the acquisition arc  started / stopped  at the first / last projection. If you refer to [Parker's article](https://www.doi.org/10.1118/1.595078), $\beta=0$ and $\beta=2\delta-\beta$ for the first and the last projections and $w=0$ in equation (12). We could instead start / stop the acquisition half an angular step before / after. This would require revisiting the angular weighting of the first / last projection.

axel-grc:
Extend the short scan range by half an angular step on each side so that the first and last projections receive non-zero Parker weights. Previously, these projections were at the ramp-up/ramp-down boundaries and received zero weight, effectively throwing them away. The acquisition actually starts half a step before the first projection and ends half a step after the last, so the weighting should reflect this.

Fix #870

### DIFF
--- a/include/rtkParkerShortScanImageFilter.hxx
+++ b/include/rtkParkerShortScanImageFilter.hxx
@@ -77,6 +77,29 @@ ParkerShortScanImageFilter<TInputImage, TOutputImage>::GenerateInputRequestedReg
   m_Delta = 0.5 * (lastAngle - m_FirstAngle - itk::Math::pi);
   m_Delta = m_Delta - 2 * itk::Math::pi * floor(m_Delta / (2 * itk::Math::pi)); // between -2*PI and 2*PI
 
+  // Extend the scan range by half a step so that the first and last projections get non-zero Parker weights.
+  if (nProj > 1)
+  {
+    // First angular step: gap from first to second projection in the short scan
+    auto itNext = itFirstAngle;
+    ++itNext;
+    double firstStep = itNext->first - itFirstAngle->first;
+    if (firstStep < 0)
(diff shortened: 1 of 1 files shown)


SimonRit (changes requested):
This is a quite technical problem to address. I'm pretty sure this won't work as there is does not fix the weight by `GetAngularGaps` in `FDKWeightProjectionFilter`. A simple test should immediately show a wrong reconstruction. It's a bit worrying that it does not show in the CI.

include/rtkParkerShortScanImageFilter.hxx
```diff
@@ -77,6 +77,18 @@ ParkerShortScanImageFilter<TInputImage, TOutputImage>::GenerateInputRequestedReg
   m_Delta = 0.5 * (lastAngle - m_FirstAngle - itk::Math::pi);
   m_Delta = m_Delta - 2 * itk::Math::pi * floor(m_Delta / (2 * itk::Math::pi)); // between -2*PI and 2*PI
 
+  // Extend scan range by half an angular step on each side.
+  // This accounts for the fact that the acquisition starts/ends
+  // half a step before/after the first/last projection, so that
+  // the first and last projections receive non-zero Parker weights.
+  if (nProj > 1)
+  {
+    double angularStep = (lastAngle - m_FirstAngle) / (nProj - 1);
```
SimonRit: I would prefer to add half the first / last angular step on each side rather than an average angular step.

PR #976, opened 2026-07-28
Various lpsolve fixes

SimonRit:


applications/rtkspectralonestep/rtkspectralonestep.sh
axel-grc, line 1: I think this is a leftover from another PR

axel-grc (commented):
I tested it and it seems to work fine.

applications/rtkspectralonestep/rtkspectralonestep.sh
SimonRit: Good catch thanks!

PR #981, opened 2026-08-19
BUG: Match CUDA forward projection clipping with CPU behaviour

### LINKED ISSUE #980: Ray may accumulate past the detector in rtkCudaForwardProjectionImageFilter.cu
RTK provides a GPU-backed forward projection image filter, which internally calls the `kernel_forwardProject` CUDA kernel defined in rtkCudaForwardProjectionImageFilter.cu. During the kernel execution, each GPU thread steps through the input volume to compute a ray integral from X-ray source to detector pixels. The start (`tnear`) and end (`tfar`) positions for the integral are defined as the entry and exit points of the ray-box intersection provided by the `intersectBox` function.

While the starting point is [explicitly clamped](https://github.com/RTKConsortium/RTK/blob/5ad0608dd50c1c1d47786890c6b8d66416c3b7af/src/rtkCudaForwardProjectionImageFilter.cu#L111) to the X-ray source location, I am surprised by the absence of a matching upper bound for `tfar`. This means that if the volume boundaries go past the detector plane, the ray integral will accumulate volume intensities beyond the detector pixel. I assume it would be more correct to also clamp the end point as:

```c++
 ray.d = pixelPos - ray.o;
 const float tmax = norm3df(ray.d);
 ray.d /= tmax;

/* compute box intersection */
// [...]

if (tnear < 0.f)
    tnear = 0.f; // clamp to near plane

if (tfar > tmax)
    tfar = tmax; // clamp to far plane
```

Conversely, the documentation of the Joseph forward projector states that "The forward projector tests if the detector has been placed after the source and the volume. If the detector is in the volume the ray tracing is performed only until that point.". Therefore, adding far-plane clamping in the GPU forward projector would match the behavior of its CPU counterpart.

maxime-y-monin:
Fixes #980


src/rtkCudaForwardProjectionImageFilter.cu
```diff
@@ -95,20 +95,19 @@ kernel_forwardProject(float * dev_proj_in, float * dev_proj_out, cudaTextureObje
     }
 
     ray.d = pixelPos - ray.o;
-    ray.d = ray.d / sqrtf(dot(ray.d, ray.d));
```
SimonRit, line 98: I'm not sure why this has been removed? It must be kept to have tnear and tfar between 0 and 1.

src/rtkCudaForwardProjectionImageFilter.cu
```diff
@@ -95,20 +95,19 @@ kernel_forwardProject(float * dev_proj_in, float * dev_proj_out, cudaTextureObje
...
-    if (!intersectBox(ray, &tnear, &tfar, c_boxMin, c_boxMax) || tfar <= 0.f || tfar == tnear)
+    if (!intersectBox(ray, &tnear, &tfar, c_boxMin, c_boxMax) || tnear >= 1.0f || tfar <= 0.f || tfar == tnear)
     {
       for (unsigned int c = 0; c < VVectorLength; c++)
         dev_proj_out[projOffset * VVectorLength + c] = dev_proj_in[projOffset * VVectorLength + c];
     }
     else
     {
-      if (tnear < 0.f)
-        tnear = 0.f; // clamp to near plane
```
SimonRit, line 111: We can replace these two lines with [__saturatef](https://docs.nvidia.com/cuda/cuda-math-api/cuda_math_api/group__CUDA__MATH__INTRINSIC__SINGLE.html#group__cuda__math__intrinsic__single_1ga2c84f08e0db7117a14509d21c3aec04e), sorry I didn't notice it initially.

SimonRit (changes requested):
Thanks for the work, this is great. In addition to my comments below, can you also add a note in `documentation/docs/rtk_3_migration_guide.md` to indicate the change of behavior. I would also squash the changes in one commit and call this an enhancement (change of behavior) rather than a bug but it's probably a matter of opinion.

src/rtkCudaForwardProjectionImageFilter.cu
maxime-y-monin: We could keep it but then the maximum value of `tfar` would be the ray direction vector magnitude (which I called `tmax` in the issue), not 1 as you proposed.

The `intersectBox` function computes scalar values `t` for line points satisfying the ray equation: `p(t) = ray.o + t * ray.d`. If `ray.d` is the segment vector from `ray.o` to `pixelPos`, then `t` would be expressed in normalized coordinates (from 0 to 1). If you normalize `ray.d` to unit magnitude, then `t` should be denormalized.

src/rtkCudaForwardProjectionImageFilter.cu
maxime-y-monin: Done in 9548ff79f73982b36ce305ee4ad6ca39b2f00381

src/rtkCudaForwardProjectionImageFilter.cu
SimonRit: Thanks for re-explaining all the code. Makes sense, sorry for the noise.

src/rtkCudaForwardProjectionImageFilter.cu
SimonRit: I now think there is no computational benefit and the new code is less readable. We can leave it as is or roll back to your initial version.

SimonRit (approved):
Thanks again for the work. Good as is but we need to update `rtkCudaRayCastBackProjectionImageFilter.cu` accordingly. Do you want to do it in this PR or do you want us to do it in a subsequent PR?

maxime-y-monin:
> Thanks again for the work. Good as is but we need to update `rtkCudaRayCastBackProjectionImageFilter.cu` accordingly. Do you want to do it in this PR or do you want us to do it in a subsequent PR?

I added a separate commit for the adjoint operator. It includes a regression test that passes only after the corresponding clipping changes in `rtkCudaRayCastBackProjectionImageFilter.cu`. Let me know if you think this needs further changes.

SimonRit (approved):
Thanks! Will merge when the CI's green.

PR #993, opened 2026-09-02
DOC: Make SVG diagrams visible in dark mode

### LINKED ISSUE #909: Documentation is not robust to black background
The transparent figures (e.g. scheme for the geometry) become invisible on a black background as they are black on black.

axel-grc:
Add custom CSS that inverts SVG images when the Furo theme's dark mode is active (both explicit toggle and OS auto preference), so the black-on- transparent geometry diagrams remain visible on a dark background.

Register the new _static directory in conf.py and include it in the list of copied sources so it is not cleaned up during the Sphinx doc build.

Fix #909 
 
The PR https://gitlab.in2p3.fr/simon.rit/rtk-website/-/merge_requests/3 must be merged to avoid conflict before merging this one

### DIFF
--- a/_static/custom.css
+++ b/_static/custom.css
@@ -1,0 +1,14 @@
+/* Invert SVG diagrams with transparent backgrounds in dark mode,
+   so black lines become visible on the dark page background. */
+
+/* When user explicitly selects dark mode in the Furo theme toggle */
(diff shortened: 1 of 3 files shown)


SimonRit (approved):
Seems to work well, thanks!

PR #996, opened 2026-09-08
BUG: Fix black frames with --cudacg and spatial TV 

### LINKED ISSUE #888: rtkfourdrooster return black frames when --cudacg option is on
When following the example of fourdrooster and activating the --cudacg option, the first frame of the reconstruction appears as intended but the other are black.
When using dvf, this problem disappear.

axel-grc:
Fixes #888 .

The total variation denoise sequence filter reused the paste output as the
destination for every frame following the first. On the ROOSTER main loop
iterations, that accumulator could alias the fed-back CG volume through the
CUDA buffer pool, so downstream spatial TV zeroed every frame but the first.

Accumulate all frames in place into a single, freshly allocated destination
buffer, which can never alias an upstream buffer.

### DIFF
--- a/include/rtkTotalVariationDenoiseSequenceImageFilter.hxx
+++ b/include/rtkTotalVariationDenoiseSequenceImageFilter.hxx
@@ -124,18 +124,18 @@ TotalVariationDenoiseSequenceImageFilter<TImageSequence>::GenerateData()
 {
   int Dimension = TImageSequence::ImageDimension;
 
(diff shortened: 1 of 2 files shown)


SimonRit (changes requested):
`rtk::TotalVariationDenoiseSequenceImageFilter` is not `InPlace`. Maybe setting the `PasteImageFilter` to `InPlaceOff` would have solved the issue more clearly?

PR #998, opened 2026-09-10
ENH: Check rotation axis is along Y in FDK reconstruction

### LINKED ISSUE #335: FDK reconstruction does not check the ramp filter direction
We should check that the orientation of the rotation axis is as expected, i.e., along the first direction of the projections.

axel-grc:
Fix #335

### DIFF
--- a/include/rtkFDKConeBeamReconstructionFilter.hxx
+++ b/include/rtkFDKConeBeamReconstructionFilter.hxx
@@ -22,6 +22,8 @@
 
 #include <itkProgressAccumulator.h>
 
+#include <cmath>
+
 namespace rtk
 {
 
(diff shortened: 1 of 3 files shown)


include/rtkFDKConeBeamReconstructionFilter.hxx
```diff
@@ -63,6 +63,19 @@ FDKConeBeamReconstructionFilter<TInputImage, TOutputImage, TFFTPrecision>::Verif
...
   if (this->m_Geometry.IsNull())
     itkExceptionMacro(<< "Geometry has not been set.");
+
+  // Check that the rotation axis is along Y (index 1), as required by the ramp filter
+  // which always applies along the first direction of the projection images.
+  const auto & matrices = this->m_Geometry->GetRotationMatrices();
+  for (unsigned int i = 0; i < matrices.size(); i++)
+  {
+    if (std::abs(matrices[i][1][0]) > 1e-6 || std::abs(matrices[i][1][1] - 1.) > 1e-6 ||
+        std::abs(matrices[i][1][2]) > 1e-6)
```
SimonRit: Too strong, we want to allow some misalignment. I would suggest checking the dot product between this second column vector and (0,1,0) and to checking it's not below a certain threshold, e.g. sqrt(3)/2 (i.e., cos(pi)/6)

SimonRit (changes requested):
Thanks for taking care of this. I'd suggest to allow some (possibly configurable) tolerance

SimonRit (approved):
Thanks for your efforts! Can be merged when the CI will be green.

include/rtkFDKConeBeamReconstructionFilter.hxx
```diff
@@ -63,6 +65,27 @@ FDKConeBeamReconstructionFilter<TInputImage, TOutputImage, TFFTPrecision>::Verif
...
+  // Check that the rotation axis is along Y (index 1), as required by the ramp filter
+  // which always applies along the first direction of the projection images. Some
+  // misalignment is allowed, up to the configurable angular tolerance.
+  const double cosTolerance = std::cos(this->m_AngularTolerance * itk::Math::pi / 180.);
+  const auto & matrices = this->m_Geometry->GetRotationMatrices();
+  for (unsigned int i = 0; i < matrices.size(); i++)
+  {
+    // Dot product between the second column vector of the rotation matrix and (0,1,0),
+    // normalized to obtain the cosine of the angle with the Y axis
+    const double norm = std::sqrt(matrices[i][1][0] * matrices[i][1][0] + matrices[i][1][1] * matrices[i][1][1] +
```
SimonRit: Each column of a rotation matrix is a unit vector I believe so norm can be dropped I think.

axel-grc: > Each column of a rotation matrix is a unit vector I believe so norm can be dropped I think.

Right, I fixed it, thanks !

SimonRit (approved):
Thanks for the fix

